PR #12, opened 2015-03-12
NUTCH-1957 using MD5 as part of file path to solve filename collision issue

renxiawang:
https://issues.apache.org/jira/browse/NUTCH-1957

Based on the comment by Sebastian Nagel on this issue, instead of storing files into "baseName"."extension", now storing files in /\<1st char of md5\>\<9th char of md5>/\<16th char of md5\>\<24th char of md5\>/\<md5\>_\<baseName\>.\<extension\>.

md5 is the MD5 sum of the file url. So given a url with MD5 sum 3f74a26653f33ea32c23d96236d6859a, with base name 24632121 and extension html, the output full path will be: 35/23/3f74a26653f33ea32c23d96236d6859a_24632121.html

If the base name is longer than 32 characters, then it will be truncated to 32 characters to avoid file name too long issue.

### DIFF
--- a/src/java/org/apache/nutch/tools/FileDumper.java
+++ b/src/java/org/apache/nutch/tools/FileDumper.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.security.MessageDigest;
 
+import com.google.common.base.Strings;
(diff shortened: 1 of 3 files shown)


chrismattmann:
Thanks committed! @renxiawang

PR #18, opened 2015-04-15
NUTCH-1987 - Make bin/crawl indexer agnostic

MJJoyce:
- Add solr.server.url property to nutch-default and set to value
  consistent with URL used in the Nutch Tutorial.
- Change SOLRURL references to INDEXFLAG for consistency.
- Update all occurrences of crawl "usage" strings to no longer reference
  solrURL and instead mention an optional string "run_indexer".
- Update indexer section to no longer set Solr URL property and remove
  Solr references from prints.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -1540,6 +1540,14 @@
 </property>
 
 <!-- solr index properties -->
+<property>
+  <name>solr.server.url</name>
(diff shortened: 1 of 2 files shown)


chrismattmann:
@MJJoyce see comments from Seb on NUTCH-1987

PR #36, opened 2015-06-24
NUTCH-2038

asitang:
Made aesthetic changes suggested by Chris Mattmann. Removed dependencies from the main ivy.xml and added it to plugin's ivy.xml.


chrismattmann:
+1 from me. I think renaming it to parserfilter-naivebayes would be fine and shorten it. Either way I am +1.

PR #50, opened 2015-08-24
[DO NOT MERGE/DISCUSSION] add cleaned up version of momer's protocol-selenium plugin

eivindveg:
Hi,

For some time, I have been using @momer's Selenium plugin for Nutch 2.3, which has worked wonders in AJAX crawling. Lately however, I've noticed the following issues with it:

1. It does not support HTTPS. While adding support for HTTPS into it, I realized that:
2. It does a lot of non-Selenium stuff that makes it inefficient, including making every request twice.

I've cleaned up the code(unfortunately, the update history pointing to him is gone because of the lazy copy), making sure it uses Selenium and only selenium.

So far it has the following weaknesses:

* Selenium's Wait does not appear to be properly used(also an issue in @momer's code)
* It does not fill out the WebPage fields according to snuff. I do think someone more experienced with Selenium could get this done, or I might add support in the future.
* Does not fetch robots.txt properly.

Due to these issues, I think merging is a bit premature. However, I'd like to point out that I do not think @momer's code should be merged either.


momer:
Very nice work! I should point out that this was a quick and dirty solution I'd cobbled together, and didn't expect at the time that it would gain much attention or use.

Thank you for cleaning it up - a great service to everyone who wants to use it!

Mo

eivindveg:
I'd like to point out that I'm not in any way holding any grudges towards your code, or yourself. You've been open about your copy pastas, and I do recall you saying somewhere you're not experienced with Java. I apologize if this pull request seems rude, it's in no way intended as such.

eivindveg:
Also, @momer, I'll write up a pull request targetting your repos if you want.

chrismattmann:
@momer's code is in the apache repos now and cleaned up so not sure what you are targeting here @evindiveg

chrismattmann:
Sorry I mean @eivindveg

momer:
No worries - but I do feel like this has contributed to many successful projects; to be fair, I spent 3 days at a coffee shop reading Core Java 1 to get familiar and write it, so the quality isn't great, or poor even, it's pretty bad admittedly. 

I'd rather have it be dependable before its used wide.

lewismc:
The PR is for Nutch 2.X codebase and also needs some work. Initial comments on my behalf include
 * dependencies are included within ivy/ivy.xml as oppose to plugin level
 * plugin 'clean' target needs to be added to src/plugin/build.xml
 * Apache License v2.0 headers need to be added throughout
 * generally speaking the entire plugin should be re-based to reflect the additional functuonality which has been committed and is being used within Nutch trunk. If you are interested in working on this @eivindveg  then great. I will be more than happy to review again.

momer:
@lewismc @eivindveg please do repair and fix the repo if you can and have the interest to. Unfortunately I overestimated my need to contribute to Nutch as a result of work. My apologies.

momer:
Note: we lost the client who relied upon the solution which takes my free time away to learn that which others rely on.

eivindveg:
Will review trunk and pull this up to line.

eivindveg:
http://svn.apache.org/repos/asf/nutch/trunk/src/plugin/protocol-selenium/src/java/org/apache/nutch/protocol/selenium/HttpResponse.java

This seems to be the newest one in trunk, and still opens a socket prior to opening the link in Selenium. Could you please point me to the cleaned up version so I have something to base things off?

lewismc:
@eivindveg 
bq. This seems to be the newest one in trunk, and still opens a socket prior to opening the link in Selenium. Could you please point me to the cleaned up version so I have something to base things off?
You are correct, that is the most recent version and your proposal would be an improvement. I think it is an excellent improvement. The code can be seen in both protocol-selenium and lib-selenium.

eivindveg:
So, I'll review your points above, and attempt to leverage Selenium to fill out the fields and account for fetcher configuration. We're currently approaching a deadline, so I'm unsure exactly when I can revisit this except for items we need ourselves; otherwise I hope to spend some time on it this weekend.

lewismc:
Please let me know where I can be of assistance. if you want to try and explain what your trying to do then I can possibly bake it in to 1.X trunk. We can then test it out. wdyt?

eivindveg:
Consider those commits me being interested. I'll work my way through quality for protocol-selenium, before starting with selenium-interactive then determining what needs to be pulled back out to lib-selenium to support that. Once both of those are compatible, I'll try some creative cherrypicking towards trunk.

eivindveg:
I've quite flamboyantly added the configuration key "http.min.render", which is described as the minimum render time before the plugin returns a html body, in milliseconds. I added this because the goal of getting the rendered links and content might sometime take longer than expected, and also because there's no time period specified before fetching otherwise(it does feel like you risk just getting the exact same non-rendered body).

Let me know if I should partially revert any parts of that.

chrismattmann:
No worries @momer you did great as a start and we constantly improve things. Thanks @eivindveg for working on this more you too @lewismc

src/plugin/protocol-selenium/src/java/org/apache/nutch/protocol/selenium/HttpResponse.java
```diff
@@ -0,0 +1,107 @@
...
+     * The nutch configuration
+     */
+    private Configuration conf = null;
+
+    public HttpResponse(Http http, URL url, WebPage page, Configuration conf) throws UnsupportedEncodingException {
+
+        this.conf = conf;
+        this.http = http;
+        this.url = url;
+
```
eivindveg: TODO: Use a HTTPUrlConnection to make a HEAD request querying for content type before opening in Selenium?

Should we attempt to load another fetcher such as protocol-httpclient if the file should not be handled by Selenium?

eivindveg:
It occurs to me that lib-selenium is alot more powerful in the merged implementation than I thought. I'll refocus on importing instead of writing code.

eivindveg:
Closing this to start over based on the trunk version.

PR #55, opened 2015-09-11
WARC exporter for the CommonCrawlDataDumper

jorgelbg:
This adds the possibility of exporting the nutch segments to a WARC files. 

From the usage point of view a couple of new command line options are available: 

* `-warc`: enables the functionality to export into WARC files, if not specified the default JACKSON formatter is used.
* `-warcSize`: enable the option to define a max file size for each WARC file, if not specified a default of 1GB per file is used as recommended by the WARC ISO standard.

The usual `-gzip` flag can be used to enable compression on the WARC files, which allow to compress the output files. 

Some changes to the default CommonCrawlDataDumper were done, essentially some changes to the Factory and to the Formats. This changes avoid creating a new instance of a CommmonCrawlFormat on each URL read from the segments.

### DIFF
--- a/src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java
+++ b/src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java
@@ -1,0 +1,286 @@
+package org.apache.nutch.tools;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.text.ParseException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.ibm.icu.text.SimpleDateFormat;
+import org.apache.commons.lang.NotImplementedException;
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.nutch.metadata.Metadata;
+import org.apache.nutch.parse.ParseData;
+
+import org.apache.nutch.parse.ParseSegment;
+import org.apache.nutch.protocol.Content;
+import org.archive.format.warc.WARCConstants;
+import org.archive.io.WriterPoolMember;
+import org.archive.io.warc.WARCRecordInfo;
+import org.archive.io.warc.WARCWriter;
+import org.archive.io.warc.WARCWriterPoolSettingsData;
+import org.archive.uid.UUIDGenerator;
+import org.archive.util.DateUtils;
+import org.archive.util.anvl.ANVLRecord;
+
+public class CommonCrawlFormatWARC extends AbstractCommonCrawlFormat {
+
+  public static final String MAX_WARC_FILE_SIZE = "warc.file.size.max";
+  public static final String TEMPLATE = "${prefix}-${timestamp17}-${serialno}";
+
+  private static final AtomicInteger SERIALNO = new AtomicInteger();
+  private final static UUIDGenerator GENERATOR = new UUIDGenerator();
+
+  private String outputDir = null;
+  private ByteArrayOutputStream out;
+  private WARCWriter writer;
+  private ParseData parseData;
+
+  public CommonCrawlFormatWARC(Configuration nutchConf,
+      CommonCrawlConfig config) throws IOException {
+    super(null, null, null, nutchConf, config);
+
+    this.out = new ByteArrayOutputStream();
+
+    ANVLRecord info = WARCUtils.getWARCInfoContent(nutchConf);
+    List<String> md = Collections.singletonList(info.toString());
+
+    this.outputDir = config.getOutputDir();
+
+    if (null == outputDir) {
+      String message = "Missing output directory configuration: " + outputDir;
+
+      throw new RuntimeException(message);
+    }
+
+    File file = new File(outputDir);
+
+    long maxSize = WARCConstants.DEFAULT_MAX_WARC_FILE_SIZE;
+
+    if (config.getWarcSize() > 0) {
+      maxSize = config.getWarcSize();
+    }
+
+    WARCWriterPoolSettingsData settings = new WARCWriterPoolSettingsData(
+        WriterPoolMember.DEFAULT_PREFIX, TEMPLATE, maxSize,
+        config.isCompressed(), Arrays.asList(new File[] { file }), md,
+        new UUIDGenerator());
+
+    writer = new WARCWriter(SERIALNO, settings);
+  }
+
+  public CommonCrawlFormatWARC(String url, Content content, Metadata metadata,
+      Configuration nutchConf, CommonCrawlConfig config, ParseData parseData)
+      throws IOException {
+    super(url, content, metadata, nutchConf, config);
+
+    this.out = new ByteArrayOutputStream();
+    this.parseData = parseData;
+
+    ANVLRecord info = WARCUtils.getWARCInfoContent(conf);
+    List<String> md = Collections.singletonList(info.toString());
+
+    this.outputDir = config.getOutputDir();
+
+    if (null == outputDir) {
+      String message = "Missing output directory configuration: " + outputDir;
+
+      throw new RuntimeException(message);
+    }
+
+    File file = new File(outputDir);
+
+    long maxSize = WARCConstants.DEFAULT_MAX_WARC_FILE_SIZE;
+
+    if (config.getWarcSize() > 0) {
+      maxSize = config.getWarcSize();
+    }
+
+    WARCWriterPoolSettingsData settings = new WARCWriterPoolSettingsData(
+        WriterPoolMember.DEFAULT_PREFIX, TEMPLATE, maxSize,
+        config.isCompressed(), Arrays.asList(new File[] { file }), md,
+        new UUIDGenerator());
+
+    writer = new WARCWriter(SERIALNO, settings);
+  }
+
+  public String getJsonData(String url, Content content, Metadata metadata,
+      ParseData parseData) throws IOException {
+    this.url = url;
+    this.content = content;
+    this.metadata = metadata;
+    this.parseData = parseData;
+
+    return this.getJsonData();
+  }
+
+  @Override
+  public String getJsonData() throws IOException {
+
+    long position = writer.getPosition();
+
+    try {
+      // See if we need to open a new file because we've exceeded maxBytes
+
+      // checkSize will open a new file if we exceeded the maxBytes setting
+      writer.checkSize();
+
+      if (writer.getPosition() != position) {
+        // We just closed the file because it was larger than maxBytes.
+        position = writer.getPosition();
+      }
+
+      // response record
+      URI id = writeResponse();
+
+      if (StringUtils.isNotBlank(metadata.get("_request_"))) {
+        // write the request method if any request info is found
+        writeRequest(id);
+      }
+    } catch (IOException e) {
+      // Launch the corresponding IO error
+      throw e;
+    } catch (ParseException e) {
+      // do nothing, as we can't establish a valid WARC-Date for this record
+      // lets skip it altogether
+      LOG.error("Can't get a valid date from: {}", url);
+    }
+
+    return null;
+  }
+
+  protected URI writeResponse() throws IOException, ParseException {
+    WARCRecordInfo record = new WARCRecordInfo();
+
+    record.setType(WARCConstants.WARCRecordType.response);
+    record.setUrl(getUrl());
+
+    String fetchTime;
+
+    record.setCreate14DigitDate(DateUtils
+        .getLog14Date(Long.parseLong(metadata.get("nutch.fetch.time"))));
+    record.setMimetype(WARCConstants.HTTP_RESPONSE_MIMETYPE);
+    record.setRecordId(GENERATOR.getRecordID());
+
+    String IP = getResponseAddress();
+
+    if (StringUtils.isNotBlank(IP))
+      record.addExtraHeader(WARCConstants.HEADER_KEY_IP, IP);
+
+    if (ParseSegment.isTruncated(content))
+      record.addExtraHeader(WARCConstants.HEADER_KEY_TRUNCATED, "unspecified");
+
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+    String httpHeaders = metadata.get("_response.headers_");
+
+    if (StringUtils.isNotBlank(httpHeaders)) {
+      output.write(httpHeaders.getBytes());
+    } else {
+      // change the record type to resource as we not have information about
+      // the headers
+      record.setType(WARCConstants.WARCRecordType.resource);
+      record.setMimetype(content.getContentType());
+    }
+
+    output.write(getResponseContent().getBytes());
+
+    record.setContentLength(output.size());
+    record.setContentStream(new ByteArrayInputStream(output.toByteArray()));
+
+    if (output.size() > 0) {
+      // avoid generating a 0 sized record, as the webarchive library will
+      // complain about it
+      writer.writeRecord(record);
+    }
+
+    return record.getRecordId();
+  }
+
+  protected URI writeRequest(URI id) throws IOException, ParseException {
+    WARCRecordInfo record = new WARCRecordInfo();
+
+    record.setType(WARCConstants.WARCRecordType.request);
+    record.setUrl(getUrl());
+    record.setCreate14DigitDate(DateUtils
+        .getLog14Date(Long.parseLong(metadata.get("nutch.fetch.time"))));
+    record.setMimetype(WARCConstants.HTTP_REQUEST_MIMETYPE);
+    record.setRecordId(GENERATOR.getRecordID());
+
+    if (id != null) {
+      ANVLRecord headers = new ANVLRecord();
+      headers.addLabelValue(WARCConstants.HEADER_KEY_CONCURRENT_TO,
+          '<' + id.toString() + '>');
+      record.setExtraHeaders(headers);
+    }
+
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+    output.write(metadata.get("_request_").getBytes());
+    record.setContentLength(output.size());
+    record.setContentStream(new ByteArrayInputStream(output.toByteArray()));
+
+    writer.writeRecord(record);
+
+    return record.getRecordId();
+  }
+
+  @Override
+  protected String generateJson() throws IOException {
+    return null;
+  }
+
+  @Override
+  protected void writeKeyValue(String key, String value) throws IOException {
+    throw new NotImplementedException();
+  }
+
+  @Override
+  protected void writeKeyNull(String key) throws IOException {
+    throw new NotImplementedException();
+  }
+
+  @Override
+  protected void startArray(String key, boolean nested, boolean newline)
+      throws IOException {
+    throw new NotImplementedException();
+  }
+
+  @Override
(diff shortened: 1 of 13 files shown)


chrismattmann:
@jorgelbg hey why did you close this?

jorgelbg:
@chrismattmann Actually I created the PR initially by mistake, but I've just reopened so we're good to go. Sorry about the confusion :smile:

jnioche:
@jorgelbg could you squash your commits?

jorgelbg:
I think we can better submit a JIRA and close this as suggested by @jnioche ?

jnioche:
that's fine. let's leave it as a PR

jnioche:
@jorgelbg  before you squash the commits, could you please fix the small issue below?
```
compile-core:
    [javac] Compiling 208 source files to /data/nutch-dipe/build/classes
    [javac] warning: [options] bootstrap class path not set in conjunction with -source 1.7
    [javac] /data/nutch-dipe/src/java/org/apache/nutch/tools/CommonCrawlFormatFactory.java:64: error: cannot find symbol
    [javac] 			return new CommonCrawlFormatWARC(nutchConf, config);
    [javac] 			           ^
    [javac]   symbol:   class CommonCrawlFormatWARC
    [javac]   location: class CommonCrawlFormatFactory
    [javac] Note: /data/nutch-dipe/src/java/org/apache/nutch/crawl/CrawlDbReader.java uses unchecked or unsafe operations.
    [javac] Note: Recompile with -Xlint:unchecked for details.
    [javac] 1 error
    [javac] 1 warning
```

jorgelbg:
Fixed, looks like the the rename of the file got lost on the commit. I'm going to squash the commits then.

chrismattmann:
awesome thanks @jorgelbg . Great work. So this PR doesn't preclude having a JIRA and isn't mutually exclusive with it. Can you:

1. Create a JIRA for this PR?
2. Reference this PR in the JIRA? 

Then we can use the process described at http://github.com/apache/nutch/#contributing to commit it and I'd be happy to help. Cheers.

jorgelbg:
I created the JIRA [NUTCH-2095](https://issues.apache.org/jira/browse/NUTCH-2095)

jnioche:
@jorgelbg, am getting

```
./nutch org.apache.nutch.tools.CommonCrawlDataDumper -segment /data/nutch-dipe/crawl1k/segments/20150817184544/ -outputDir  /data/nutch-dipe/crawl1k/warc/20150817184544/ -warc
java.lang.RuntimeException: Missing output directory configuration: indexer.warc.output
```

Shouldn't that value be taken from the value of  -outputDir or are they meant to be different?

jnioche:
While we are at it, could you change the CCDD class so that it implements Tool?  It currently chokes on 
`/nutch org.apache.nutch.tools.CommonCrawlDataDumper -D indexer.warc.output=/data/nutch-dipe/crawl1k/warc/20150817184544/ `
Always nicer to be able to pass config on the command line

jnioche:
@jorgelbg would also be good to pass either an individual segment or the whole directory containing the segments as we do for other tools. The name of the parameter is currently misleading (`-segment`) as in fact it considers the parent segments directory.

Also we should filter the content metadata that do not come from http response header i.e. the ones starting with _ or nutch.
Note that WARC-DATE should never be null as it is a mandatory value. If we can't generate a value at the correct format then the entry should be skipped altogether.

```
WARC/1.0
WARC-Type: response
WARC-Target-URI: http://11st.co.kr/
WARC-Date: null
WARC-IP-Address: 
WARC-Record-ID: <urn:uuid:9b4bd057-c361-4425-a5f8-cd4cd121665e>
Content-Type: application/http; msgtype=response
Content-Length: 967

Status: 
Hostname: 11st.co.kr
Address: 
Content-Encoding: 
Content-Type: text/html; charset=EUC-KR
Date: Mon, 17 Aug 2015 15:58:55 GMT
Server: Apache
_depth_: 1
Vary: User-Agent
Date: Mon, 17 Aug 2015 15:58:55 GMT
Content-Length: 369
_fst_: 33
nutch.segment.name: 20150817165846
Set-Cookie: JSESSIONID=GzspVSGfM2VPp4r8yNGM92RdJHmLRdm72Tjf1nFJWJJhLrTsn7xy!1924126666!-1408058590!9604!-1; path=/
_maxdepth_: 3
Content-Type: text/html; charset=EUC-KR
Connection: close
X-Powered-By: Servlet/2.4 JSP/2.0
Server: Apache
Cache-Control: no-cache="set-cookie"
```

One option is to return a resource instead of a response (i.e no http headers included) but I expect that it is still be better to have some noise in the headers than no http headers at all.

Are the http headers included in the Content-Length: ?

Could we also replace the spaces with %20 instead of skipping the entries that do not comply for WARC-Target-URI?

Thanks!

jnioche:
Also it would be good to be able to set a max size for the content and trim it accordingly. There is a special keyword in the WARC specs that allows you to declare that a content has been trimmed. Am getting memory exceptions when running the tool and I'm wondering whether it would make it more robust.

Plus it  would be good to detect whether the content had been trimmed while fetching and use that flag in such cases (see [https://github.com/DigitalPebble/nutch/blob/warcjn/src/java/org/apache/nutch/tools/warc/WARCExporter.java#L177]) for an example.

lewismc:
+1 good examples

On Saturday, September 12, 2015, Julien Nioche <notifications@github.com>
wrote:

> Also it would be good to be able to set a max size for the content and
> trim it accordingly. There is a special keyword in the WARC specs that
> allows you to declare that a content has been trimmed. Am getting memory
> exceptions when running the tool and I'm wondering whether it would make it
> more robust.
>
> Plus it would be good to detect whether the content had been trimmed while
> fetching and use that flag in such cases (see [
> https://github.com/DigitalPebble/nutch/blob/warcjn/src/java/org/apache/nutch/tools/warc/WARCExporter.java#L177])
> for an example.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/nutch/pull/55#issuecomment-139748971>.
>


-- 
*Lewis*

jorgelbg:
I'm working on Julien suggestions will commit soon.

chrismattmann:
great work and +1 to @jnioche comments. @giuseppetotaro please also look at this. thanks @jorgelbg

giuseppetotaro:
Great work. Thanks @chrismattmann for bringing it to my attention. I will be on this as soon as possible.

jnioche:
We could also store the http headers as-is in the content metadata by adding a config like `http.store.httpheaders`; if we find a value under this key then we could use that for returning the response if not we could try and rebuild one by removing the key values added by Nutch as suggested above. This would guarantee a more accurate response entity in the WARC.

lewismc:
Nice

On Sunday, September 13, 2015, Julien Nioche <notifications@github.com>
wrote:

> We could also store the http headers as-is in the content metadata by
> adding a config like http.store.httpheaders; if we find a value under
> this key then we could use that for returning the response if not we could
> try and rebuild one by removing the key values added by Nutch as suggested
> above. This would guarantee a more accurate response entity in the WARC.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/nutch/pull/55#issuecomment-139849360>.
>


-- 
*Lewis*

jorgelbg:
@jnioche Perhaps the same approach can be used to store data about the request made by Nutch and populate a better request record.

jnioche:
Yes good idea

jorgelbg:
@jnioche The OOM exception is raised when writing to the WARC file? Should we use the same `*.content.limit` configuration potion or use a CLI parameter instead?

jorgelbg:
Also with the last commit the tool is able of reading a single segment or the "parent" directory with all the segments, should I add also the option of specifying several segments? I think this could be useful.

lewismc:
Yes I think it would be useful indeed.


On Monday, September 14, 2015, Jorge Luis Betancourt <
notifications@github.com> wrote:

> Also with the last commit the tool is able of reading a single segment or
> the "parent" directory with all the segments, should I add also the option
> of specifying several segments? I think this could be useful.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/nutch/pull/55#issuecomment-140087239>.
>


-- 
*Lewis*

jnioche:
> The OOM exception is raised when writing to the WARC file?

yes + this is mentioned on the WIKI page for the CCDD

> Should we use the same *.content.limit configuration potion or use a CLI parameter instead?

I'd rather have a separate config option. no need to add it to the already long list of CLI parameters I think

jnioche:
> Also with the last commit the CCDD is able of reading a single segment or the "parent" directory with all the segments, should I add also the option of specifying several segments?

Why not but I'd rather have a consistent behaviour across our Tools and later on add that functionality to all of them - maybe make so that they use the same code for that

src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java
```diff
@@ -0,0 +1,337 @@
...
+      if (writer.getPosition() != position) {
+        // We just closed the file because it was larger than maxBytes.
+        position = writer.getPosition();
+      }
+
+      // response record
+      URI id = writeResponse();
+
+      // request record
+      writeRequest(id);
```
jnioche: add a config so that can decide whether to handle the requests?

src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java
```diff
@@ -0,0 +1,337 @@
...
+
+      // metadata record
+      writeMetadata(id);
```
jnioche: same - add a config for this?

jnioche:
Hi @jorgelbg. This is not ready to be committed yet as it generates incorrect HTTP headers.
You can use [https://bitbucket.org/nclarkekb/jwat-tools-gui/downloads] to verify the validity of the WARC files. [https://pypi.python.org/pypi/Warcat/] is also useful.

Am currently getting 

```
Status: 
Hostname: www.tescobank.com
Content-Encoding: gzip
Content-Type: text/html;charset=UTF-8
Date: Fri, 04 Sep 2015 14:26:58 GMT
Server: Tesco Bank
Server: Tesco Bank
Content-Encoding: gzip
Connection: close
Vary: Accept-Encoding
Set-Cookie: JSESSIONID=B9DDE1C6C396DD933E141842BA7426FF.node1; Path=/;HttpOnly
partialRefresh: true
Content-Type: text/html;charset=UTF-8
```
Note the double `Server` entry 

It should start with something like `HTTP/1.1 200 OK`. Could you clarify where that `Status` value comes from?

This reinforces my suggestion that we should capture the HTTP header in the low level HTTP response class instead of trying to regenerate it from the metadata. It would also preserve the real order in which headers were returned.

src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java
jorgelbg: :+1:

src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java
jorgelbg: Yes actually I was thinking this for the case that we decided to read the ParseData as well to add this into a metadata record.

jnioche: Let's keep it simple for now, we can always add that later on

jnioche:
See [https://github.com/DigitalPebble/nutch/commit/56bc80c35f152710e62a417ad60b37b501387b03] for storing request and response headers verbatim.

jorgelbg:
Missed that duplicated Server header. The status code is built from the `Metadata`, but with the option to store the raw request/response this becomes unnecessary (but configurable). You beat me to storing the raw request/response :) are you going to create a separated PR or should I incorporate those changes into this PR? Also we're going to keep this manual header processing as a backup option?

jnioche:
> Are you going to create a separated PR or should I incorporate those changes into this PR?
Please add to this PR.

> Also we're going to keep this manual header processing as a backup option?
No, let's return an WARC element of type 'resource' if no headers have been found, much cleaner I think and still very useful.

jnioche:
Also we shouldn't generate the Date from the one taken from the Http headers [https://github.com/DigitalPebble/nutch/blob/warc/src/java/org/apache/nutch/tools/CommonCrawlFormatWARC.java#L167] but maybe add a clean new metadata at the right format within the Fetcher e.g. `nutch.fetch.date`, unless we already have something like this somewhere.

jorgelbg:
I have no problem in defining this metadata for the fetch time, I was kind of guiding by the explanation of the Date header in the HTTP RFC. But truth be told, is not very good to rely on this as you said.

jorgelbg:
Also in the case that no `_request_` key is found in the Metadata object we should skip and generate only the response record right?

jnioche:
> Also in the case that no _request_ key is found in the Metadata object we should skip and generate only the response record right?

yes

jorgelbg:
Perhaps `CrawlDatum.fetchTime` could be used, but since this can have the date of the future crawl date of an URL see received https://github.com/apache/nutch/blob/trunk/src/java/org/apache/nutch/crawl/CrawlDatum.java#L178-L184 But the index-basic plugin rely on this property for indexing the fetched time, so any comments on this?

jnioche:
I assume the value found in crawl_fetch would be correct, let's use that

jnioche:
Ok chaps, I've just opened [https://issues.apache.org/jira/browse/NUTCH-2102] which has a patch creating a WARCExporter based on MapReduce whereas the CCDD is a local file, single JVM sort of thing. As explained there, we will have a WARCImporter later on, which is outside the scope of the CCDD anyway + strictly speaking WARC is not specific to CommonCrawl so the package name I used in NUTCH-2102 is probably more appropriate.

That patch contains the changes to HttpResponse.java that we have here, whichever approach gets committed first - if any - means that we'd need to modify the other but no big deal.

Personally I'd have no problem in having some duplication of functionality between NUTCH-2012 and this improvement to the CCDD (we've done a lot worse over the years). Maybe if and when the CCDD is based on MapReduce and supports IO on DFS then we can reconsider whether to keep NUTCH-2102 or not.

chrismattmann:
Great work @jorgelbg to integrate this into the existing export framework for CCDD - and now you have a new NUTCH-2102 that is based off M/R. I am fine as long as NUTCH-2102 is a basis for turning all the jobs into M/R. I'll take a look at it. Thanks @jnioche

chrismattmann:
cc / @giuseppetotaro

jnioche:
> the nutch.fetch.time is initialized with the CrawlDatum.fetchTime property, on the HttpResponse class since now the CCDD only reads the content of the segment and doesn't read any other data. Any suggestion on this?

Just  modify the protocol so that it creates 'nutch.fetch.time' in the content medatata. Taking the Dates returned by the servers does not give any guarantee that they are accurate, let alone that they exist.

jorgelbg:
I've just squashed the changes into a single commit.

src/plugin/protocol-http/src/java/org/apache/nutch/protocol/http/HttpResponse.java
```diff
@@ -193,34 +197,54 @@ public HttpResponse(HttpBase http, URL url, CrawlDatum datum)
...
+                  Http.BUFFER_SIZE), Http.BUFFER_SIZE);
 
       StringBuffer line = new StringBuffer();
 
+      // store the http headers verbatim
+      if (conf.getBoolean("store.http.headers", false) == true) {
+        httpHeaders = new StringBuffer();
+      }
+
+      headers.add("nutch.fetch.time", Long.toString(datum.getFetchTime()));
```
jnioche: Why take `datum.getFetchTime()` and what does its value correspond to? Isn't that the time is was due for fetching?
I'd have used `System.currentTimeMillis()` as this is exactly the moment we fetch the content and would definitely be accurate

jorgelbg: Though of using the `System.currentTimeMillis()` method but reviewing this comment https://github.com/apache/nutch/pull/55#issuecomment-140663159 though that it will had the right value. 

The comment on the `getFetchTime()` method says:
 
  > Returns either the time of the last fetch, or the next fetch time,
  > depending on whether Fetcher or CrawlDbReducer set the time.

Since this is executed in the HttpResponse class (right after the fetcher gets executed) I though it would be save to assume that the date would be accurate. If this is wrong the fix is easy enough.

jnioche: It is correct in the output of the fetcher step when accessing the fetch datum but I don't think it is the case at this point in the code

> it is executed in the HttpResponse class (right after the fetcher gets executed)

not after but right in the middle of the fetcher's work. 

It is set to the right value in the output method of the fetcherthread [https://github.com/apache/nutch/blob/8397611b49de4aac408806765191fc796ba4b15f/src/java/org/apache/nutch/fetcher/FetcherThread.java#L528] but that's AFTER the protocol implementation fetched the content.

In short not clear what the value is at this point of the code but it's unlikely to be correct. Just use System.currentTimeMillis()

jorgelbg: I understand, fixed/pushed then!, the misleading portion of the comment for me was that it should hold the right value unless the `CrawlDbReducer` gets called. The `fetchTime` parameter gets initialized to the `System.currentTimeMillis()` upon creation. But this way we have 0 doubts about the value.

jnioche:
Great, thanks @jorgelbg. Am +1 to commit this.

@chrismattmann what about taking small incremental steps and commit NUTCH-2102 first then modify it so that it can handle the formats done by the CCDD?

In practice the CCDD is of little use even on medium sized crawls as it does not do HDFS and mapreduce unlike NUTCH-2102. If you think about CommonCrawl who would be the main users of the WARC exporter and given the volume of data they deal with, NUTCH-2102 is probably the option they'll go for. They already have some code for it, sure, but I'd rather they used and contributed to stuff which is in our codebase

jorgelbg:
Actually I think that @jnioche has a point in terms of scalability NUTCH-2102 has a better approach. The ideal approach could be to merge both tools into one. That will run locally and at scale (which is one of the main points of Nutch).

jorgelbg:
@chrismattmann we are good to commit this, since the branch name is not in the format defined in https://github.com/apache/nutch/#contributing there is any additional step required? Or can I commit like usual?

jnioche:
@jorgelbg what about 
`git diff --no-prefix trunk..warc > NUTCH-2095.patch`
attach it to the JIRA, patch then commit on SVN? This way the JIRA issue will have the patch and you'll be able to modify the CHANGES.txt as well

chrismattmann:
@jnioche happy with incremental steps. @jorgelbg feel free to commit as usual or attach the patch however you feel fit. Great work as usual!

chrismattmann:
One quick comment - my group extensively uses CCDD - even on medium crawls. We just dump on a big head node. Does this alter that behavior?

jnioche:
Implemented in [23c7761aff83]

@chrismattmann I don't expect that this would change anything to the existing functionalities of CCDD. @jorgelbg can you confirm?

Can we close this PR?

jorgelbg:
Yes this doesn't change the default behavior of CCDD, this only adds the option of also exporting into the WARC format. So change to the implemented logic was done. 

+1 On closing this PR.

jnioche:
> +1 On closing this PR.

@jorgelbg could you please do it? I don't have write access on GH. Thanks!

chrismattmann:
great work. Note in general when committing if you say in your commit message "this closes #<GITHUB ISSUE NUMBER>" the asfbot will close it for you on SVN commit hook. Thanks @jorgelbg @jnioche

PR #56, opened 2015-09-12
Nutch 2096: Explicitly indicate broswer binary to use when selecting selenium remote option in config

kwhitehall:
*Assumes that the selenium grid is homogenous i.e. uses the same browser on the hub and the nodes and these are located at a similar location


lewismc:
Nice patch Kim.

PR #61, opened 2015-09-21
Fix for NUTCH-2086 Contributed by Sujen Shah

sujen1412:


lewismc:
OK will scope and provide feedback ASAP.
Thanks for this what an effort!

PR #63, opened 2015-09-22
Update NutchServer.java

zhangmianhongni:
Line 201, "CMD_PORT" is incorrect, edit to "CMD_HOST".


lewismc:
Thank you @zhangmianhongni for this patch. Good catch.

PR #64, opened 2015-09-23
fix for NUTCH-2111 contributed by kwhitehall

kwhitehall:
Further investigation showed that changing the temporary path does not get rid of the tmp files that eat up space. Further, if a selenium grid is utilized, the location chosen on a given node may not be available on all nodes. As such, it is best to stay with the default /tmp location and handle deleting the files there instead. The patch submitted does this.


lewismc:
Please close this off @kwhitehall thank you

PR #70, opened 2015-10-10
Fix for NUTCH-2135 by Sujen Shah

sujen1412:


lewismc:
Hi @sujen1412 
I wonder if you could name your issues as you would add them in the CHANGES.txt?
e.g. 
```
NUTCH-2135 Ant Eclipse build does not include protocol-interactiveselenium
```
This way it is much easier for people to see what the issue relates to and whether or not they are interested in either reviewing it, or pulling it locally and testing it.
This is just a suggestion, it is by no means a requirement. Great work.

PR #78, opened 2015-10-15
Fix for NUTCH-2139 contributed by jorgelbg

jorgelbg:
Basic indexing capabilities for inlinks and outlinks.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -1896,4 +1896,37 @@ CAUTION: Set the parser.timeout to -1 or a bigger value than 30, when using this
   </description>
 </property>
 
+<!-- index-links plugin -->
+
+<property>
+  <name>index.links.outlinks.host.ignore</name>
+  <value>false</value>
+  <description>
+    Ignore outlinks that point out to the same host as the URL being indexed. 
+    By default all outlinks are indexed. If db.ignore.internal.links is true (default
+    value), this setting does nothing since the internal links are already
+    ignored.
+  </description>
+</property>
+
+<property>
+  <name>index.links.inlinks.host.ignore</name>
+  <value>false</value>
+  <description>
+    Ignore inlinks coming from the same host as the URL being indexed. By default 
+    all inlinks are indexed. If db.ignore.internal.links is true (default
+    value), this setting does nothing since the internal links are already
+    ignored.
+  </description>
+</property>
+
+<property>
+  <name>index.links.hosts.only</name>
+  <value>false</value>
+  <description>
+    This force the index-links plugin to only index the host portion of the inlinks
+    or outlinks.
+  </description>
+</property>
+
 </configuration>
--- a/src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
+++ b/src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
@@ -1,0 +1,167 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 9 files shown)


src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
```diff
@@ -0,0 +1,168 @@
...
+      LOG.error("Malformed URL in {}: {}", url, e.getMessage());
+    }
+
+    // Add the inlinks
+    if (null != inlinks) {
+      Iterator<Inlink> iterator = inlinks.iterator();
+      Set<String> inlinkHosts = new HashSet<String>();
+
+      try {
+        while (iterator.hasNext()) {
```
sebastian-nagel: Shouldn't the try-catch be inside the while loop? The first invalid URL would break the loop and cause remaining inlinks to be skipped from being indexed. Is this intended? Ok, this should never happen since inlinks are validated URLs pointing to already fetched documents.

src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    try {
+      if (outlinks != null) {
+        Set<String> hosts = new HashSet<String>();
+
+        for (Outlink outlink : outlinks) {
+          String linkUrl = outlink.getToUrl();
+          String outHost = new URL(linkUrl).getHost();
+
+          if (indexHost) {
+            linkUrl = new URL(outlink.getToUrl()).getHost();
```
sebastian-nagel: now linkurl is equal to outHost - could be simplified

src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
```diff
@@ -0,0 +1,168 @@
...
+
+  @Override
+  public NutchDocument filter(NutchDocument doc, Parse parse, Text url,
+      CrawlDatum datum, Inlinks inlinks) throws IndexingException {
+
+    // Add the outlinks
+    Outlink[] outlinks = parse.getData().getOutlinks();
+
+    try {
+      if (outlinks != null) {
```
sebastian-nagel: see comment below regarding nesting of try-catch and loops: outlinks are not necessarily validated inside parse data

conf/nutch-default.xml
```diff
@@ -1896,4 +1896,33 @@ CAUTION: Set the parser.timeout to -1 or a bigger value than 30, when using this
   </description>
 </property>
 
+<!-- index-links plugin -->
```
sebastian-nagel: by best practice, the 3 properties would share a common prefix, so that it's clear to which plugin they belong

conf/nutch-default.xml
```diff
@@ -1896,4 +1896,33 @@ CAUTION: Set the parser.timeout to -1 or a bigger value than 30, when using this
...
+  </description>
+</property>
+
+<property>
+  <name>inlinks.host.ignore</name>
+  <value>false</value>
+  <description>
+    Ignore the inlinks coming from the same host as the URL being indexed. By default 
+    all inlinks are indexed.
+  </description>
```
sebastian-nagel: If db.ignore.internal.links == true (that's the default) this option wouldn't change anything because host-internal-inlinks aren't even stored in linkdb. Could be worth to place a reference to db.ignore.internal.links.

src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
jorgelbg: I though the same, since the URL is already is fetched shouldn't be any trouble, but its an easy fix so I can put it inside the while loop.

conf/nutch-default.xml
jorgelbg: Any thoughs on a good prefix to use? `index.links.outlinks.host.ignore` seems extremely large for my taste :)

conf/nutch-default.xml
jorgelbg: Indeed.

chrismattmann:
@jorgelbg please add your name into CHANGES.txt part of your PR (after the issue ID like the other issues). Also this has merge conflicts.

jorgelbg:
@chrismattmann Is this ok?, I've fixed the merge conflicts

lewismc:
+1

chrismattmann:
+1 to merge. Please note the pull request number (this close .. PR number) in your commit message. Thanks! @jorgelbg

src/plugin/index-links/src/java/org/apache/nutch/indexer/links/LinksIndexingFilter.java
sebastian-nagel: +1 great!

PR #79, opened 2015-10-22
Fix for NUTCH-2146 contributed jorgelbg

jorgelbg:
The `Outlink` class doesn't have a `hashCode()` method. This doesn't cause any trouble with the already implemented plugins but if a developer tries to use a `HashSet` of outlinks in a custom plugin, the `Outlink` instances with same data (toUrl, anchor) gets added several times. In contrast the `Inlink` class have a hashCode method. 

https://github.com/apache/nutch/blob/trunk/src/java/org/apache/nutch/crawl/Inlink.java#L75-L77

The same logic used in the `Inlink` class is implemented in this PR.

### DIFF
--- a/src/java/org/apache/nutch/parse/Outlink.java
+++ b/src/java/org/apache/nutch/parse/Outlink.java
@@ -128,4 +128,8 @@ public class Outlink implements Writable {
     return repr.toString();
   }
 
+  @Override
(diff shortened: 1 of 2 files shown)


lewismc:
+1, tested locally tests pass.

PR #83, opened 2015-10-28
NUTCH-2155 - Add crawl completion utility

MJJoyce:
- Add simple crawl completion utility that reports count of fetch and
  unfetched pages per domain or host.
- Update "nutch" helper script with new utility command.

### DIFF
--- a/src/java/org/apache/nutch/util/CrawlCompletionStats.java
+++ b/src/java/org/apache/nutch/util/CrawlCompletionStats.java
@@ -1,0 +1,189 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nutch.util;
+
+import java.io.IOException;
+import java.net.URL;
+import java.text.SimpleDateFormat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


chrismattmann:
@MJJoyce I love this tool! thank you will scope

src/java/org/apache/nutch/util/CrawlCompletionStats.java
```diff
@@ -0,0 +1,189 @@
...
+  private static final Logger LOG = LoggerFactory
+      .getLogger(CrawlCompletionStats.class);
+
+  private static final int MODE_HOST = 1;
+  private static final int MODE_DOMAIN = 2;
+
+  private int mode = 0;
+
+  public int run(String[] args) throws Exception {
+    if (args.length < 2) {
```
lewismc: Not a absolute requirement but merely a suggestion, it would be GREAT to see commons-cli used here to prevent incorrect CLI usage. It also prints much more user friendly output when invoked without options or with '-h'.
For excellent examples of where commons-cli is already used please see [here](https://github.com/apache/nutch/blob/trunk/src/java/org/apache/nutch/scoring/webgraph/WebGraph.java#L698-L760) @MJJoyce.
I like this job as well, it's a neat and quick way to see domain coverage.

MJJoyce: +1 I agree completely @lewismc. I got a bit lazy and stole some from domainstats (which is also in need of some commons-cli love as well). I'll try to throw a patch together an address some of these issues when I get some free time.

src/java/org/apache/nutch/util/CrawlCompletionStats.java
```diff
@@ -0,0 +1,189 @@
...
+    job.setNumReduceTasks(numOfReducers);
+
+    try {
+      job.waitForCompletion(true);
+    } catch (Exception e) {
+      throw e;
+    }
+
+    long end = System.currentTimeMillis();
+    LOG.info("CrawlCompletionStats: finished at " + sdf.format(end) + ", elapsed: "
```
lewismc: We can use slf4j conventions [here](http://www.slf4j.org/faq.html#logging_performance) e.g. 
```
LOG.info("CrawlCompletionStats: finished at {}, elapsed: {}", sdf.format(end), TimingUtil.elapsedTime(start, end));
```

lewismc:
Looking good @MJJoyce

src/java/org/apache/nutch/util/CrawlCompletionStats.java
MJJoyce: Noted and updated

chrismattmann:
going to commit this now @MJJoyce @lewismc

PR #88, opened 2016-02-10
NUTCH-2213 : do not store the headers verbatim if the response was compressed

jnioche:
See discussion on [https://issues.apache.org/jira/browse/NUTCH-2213].

### DIFF
--- a/src/plugin/protocol-http/src/java/org/apache/nutch/protocol/http/HttpResponse.java
+++ b/src/plugin/protocol-http/src/java/org/apache/nutch/protocol/http/HttpResponse.java
@@ -238,10 +238,6 @@ public class HttpResponse implements Response {
         haveSeenNonContinueStatus = code != 100; // 100 is "Continue"
       }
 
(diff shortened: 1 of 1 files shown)


lewismc:
+1 @jnioche

src/plugin/protocol-http/src/java/org/apache/nutch/protocol/http/HttpResponse.java
```diff
@@ -256,6 +252,11 @@ public HttpResponse(HttpBase http, URL url, CrawlDatum datum)
       } else if ("deflate".equals(contentEncoding)) {
         content = http.processDeflateEncoded(content, url);
       } else {
+        // store the headers verbatim only if the response was not compressed
+        // as the content length reported with not match otherwise
```
MJJoyce: "with" should be "will" here I think

MJJoyce:
+1, looks great to me!

PR #91, opened 2016-02-12
NUTCH-2218 - Update CrawlComplete util with Commons CLI arg parsing

MJJoyce:
- Switch all argument parsing and checking to commons CLI.
- Update input directory processing such that the 'crawldb' folder
  should no longer be included in the input paths. Now the user simply
  points to the crawl folder instead of specific crawldb directories.

### DIFF
--- a/src/java/org/apache/nutch/util/CrawlCompletionStats.java
+++ b/src/java/org/apache/nutch/util/CrawlCompletionStats.java
@@ -23,6 +23,14 @@ import java.net.URL;
 import java.text.SimpleDateFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.GnuParser;
+import org.apache.commons.cli.HelpFormatter;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/nutch/util/CrawlCompletionStats.java
```diff
@@ -61,27 +69,60 @@
   private int mode = 0;
 
   public int run(String[] args) throws Exception {
-    if (args.length < 3) {
-      System.err.println("Usage: CrawlCompletionStats inputDirs outDir mode [numOfReducer]");
+    Option helpOpt = new Option("h", "help", false, "Show this message");
+    Option inDirs = OptionBuilder
+        .withArgName("inputDirs")
+        .isRequired()
+        .withDescription("Comma separated list of crawl directories")
```
lewismc: This could do with being explicit e.g. providing an example

lewismc:
Nice. +1 to this if you can provide an example.

MJJoyce:
+1 @lewismc I'll see about throwing an example in there when I commit.

MJJoyce:
Note, closing this manually since I buggered the merge =D

PR #99, opened 2016-03-22
NUTCH-2222 re-fetch deletes all metadata except _csh_ and _rs_

lewismc:
Hi Folks,
This pull request formalizes the contribution from Adnane B. over on [NUTCH-2222](https://issues.apache.org/jira/browse/NUTCH-2222). I am not sure that this is catching the bug however and it seems like the bug only rears it's head when we use a persistent datastore. 
Right now this patch uses the MemStore implementation which is what we agreed tests would be like. I think we need to do add to the tests in order to have a persistent datastore setup and that data is persisted into the datastore.

### DIFF
--- a/src/java/org/apache/nutch/crawl/GeneratorJob.java
+++ b/src/java/org/apache/nutch/crawl/GeneratorJob.java
@@ -240,9 +240,17 @@ public class GeneratorJob extends NutchTool implements Tool {
 
   /**
    * Mark URLs ready for fetching.
-   * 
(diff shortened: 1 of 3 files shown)


lewismc:
Anyone able to have a look at this patch? I am tempted to merge and then work on another patch which uses a persistent datastore to verify if (and why) data is being lost.

lewismc:
I am going to create a separate test which utilizes a persistent Datastore... it turns out that this is required now.

PR #101, opened 2016-03-31
Fix for NUTCH-2245 NGram Model for Cosine Similarity by bhavyasanghavi

bhavyasanghavi:


bhavyasanghavi:
Description:

1) The user should enable the plugin by enabling scoring-similarity in the plugin.includes property in nutch-site.xml.
2) Copy the gold-standard file into the conf directory and enter the name of this file in nutch-site.xml.
3) The user should include the following property in nutch-site.xml to specify the window size for ngram model:
`<property>`
    `<name>scoring.similarity.ngrams</name>`
    `<description>Specifies the 'n' in ngrams</description>`
`</property>`
4) If the user does not include the above mentioned property, the default size will be 1, i.e unigrams will be considered for similarity scoring.

Use Case: It is extremely useful in cases where the user specifies names entities in the gold standard. For example "Leonardo DiCaprio"  is one of the terms in gold standard. The ngram model with a window size 2, will provide distinction between pages pertaining to "Leonardo da Vinci", which otherwise would not be captured. Another example is where user is looking for phrases/dialogues, the window size can be cleverly adjusted depending on the content of the gold standard.

src/plugin/scoring-similarity/src/java/org/apache/nutch/scoring/similarity/cosine/Model.java
```diff
@@ -68,6 +68,11 @@ public static synchronized void createModel(Configuration conf) throws IOExcepti
         }
         LOG.info("Loaded custom stopwords from {}",conf.get("scoring.similarity.stopword.file"));
       }
+
+      //Check if user has specified n for ngram cosine model
+      int ngram = conf.getInt("scoring.similarity.ngrams", 1);
+      LOG.info("Value of ngram: "+ngram);
```
lewismc: Please use correct effficient slf4j code notation here
e.g. LOG.info("Value of ngram: {} ", ngram);

src/plugin/scoring-similarity/src/java/org/apache/nutch/scoring/similarity/util/LuceneTokenizer.java
```diff
@@ -88,6 +89,12 @@ public TokenStream getTokenStream() {
     return tokenStream;
   }
   
+  public LuceneTokenizer(String content, TokenizerType tokenizer, StemFilterType stemFilterType, int ngram) {
```
lewismc: No Javadoc?

lewismc:
Overall, the patch it looking excellent.

src/plugin/scoring-similarity/src/java/org/apache/nutch/scoring/similarity/cosine/Model.java
bhavyasanghavi: Yes, I will update that. Thanks.

src/plugin/scoring-similarity/src/java/org/apache/nutch/scoring/similarity/cosine/Model.java
```diff
@@ -115,6 +126,7 @@ public static DocVector createDocVector(String content) {
       tStream.reset();
       while(tStream.incrementToken()) {
         String term = charTermAttribute.toString();
+        LOG.info(term);
```
sujen1412: This seems like its used for debugging, please change it to LOG.debug(). It helps keeping the log clean. 
Thanks!

bhavyasanghavi: Updated it. Thanks.

sujen1412:
I have tried out the patch, works as expected, going to commit now. 
Thanks @bhavyasanghavi!

PR #102, opened 2016-04-07
NUTCH-2248 CSS parser plugin

naegelejd:
As described on JIRA:

This plugin allows for collecting uri links from CSS (stylesheets). This is useful for collecting parent stylesheets, fonts, and images needed to display web pages as intended.

Parsed Outlinks do not have associated anchors, and no additional text/content is parsed from the stylesheet.

### DIFF
--- a/src/plugin/parse-css/src/java/org/apache/nutch/parse/css/CssParser.java
+++ b/src/plugin/parse-css/src/java/org/apache/nutch/parse/css/CssParser.java
@@ -1,0 +1,225 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.nutch.parse.css;
+
+import java.io.*;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.steadystate.css.parser.CSSOMParser;
+import com.steadystate.css.parser.SACParserCSS3;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.nutch.metadata.Metadata;
+import org.apache.nutch.protocol.Content;
+import org.apache.nutch.parse.*;
+import org.apache.nutch.util.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.w3c.css.sac.CSSParseException;
+import org.w3c.css.sac.InputSource;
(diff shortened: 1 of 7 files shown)


lewismc:
Hi @naegelejd this is a cool patch. There is unfortunately a pretty serious issue. Unfortunately, we cannot release this plugin within Apache software. Please see http://www.apache.org/licenses/GPL-compatibility.html, the thirst party CSS parser code is LGPL licensed... which is a real shame.

lewismc:
@naegelejd maybe you could use https://github.com/corgrath/osbcp-css-parser???

naegelejd:
Thanks @lewismc, I admit I actually didn't check the license of the third-party CSS parser library. It's unfortunate that this is an issue at all, especially since Nutch plugins are just that: optional plugins. I don't think I want to switch to the other library you mentioned as the LGPL licensed one is better maintained and implements the [SAC](https://www.w3.org/Style/CSS/SAC/Overview.en.html).

 I noticed you brought up a plugin central discussion on nutch-dev. Perhaps this plugin would be better hosted on Maven Central, available to Nutch at build-time. Let me know if I can help push that forward.

lewismc:
A agree with everything your saying. I think we need to address this plugin issue and host everything on Maven central. Lets keep this issue open for the time being. The plugin looks grand. I am going to propose a couple of suggestions so feel free to take them on board if you have time.

src/plugin/parse-css/src/java/org/apache/nutch/parse/css/CssParser.java
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Suppresses warnings, logs all errors, and throws fatal errors
+   */
+  private class ErrorHandler implements org.w3c.css.sac.ErrorHandler {
+    @Override
+    public void warning(CSSParseException exception) { }
+
+    @Override
+    public void error(CSSParseException exception) {
+      LOG.debug("CSS parser error: " + exception.getMessage());
```
lewismc: Can you use ```LOG.debug("CSS parser error: {}", exception.getMessage());```

src/plugin/parse-css/src/java/org/apache/nutch/parse/css/CssParser.java
```diff
@@ -0,0 +1,225 @@
...
+
+    // resolve each relative URL to create a list of Outlinks
+    List<Outlink> outlinks = new ArrayList<Outlink>();
+    for (int i = 0; i < urls.size(); i++) {
+      String rawUrl = urls.get(i);
+      try {
+        URL url = URLUtil.resolveURL(base, rawUrl);
+        outlinks.add(new Outlink(url.toString(), ""));
+      } catch (MalformedURLException e) {
+        LOG.debug("failed to resolve url (base: " + base + ", path: " + rawUrl + ")");
```
lewismc: Can you use ```LOG.debug("failed to resolve url (base: {}, path: {})", base, rawUrl); ```

naegelejd:
Of course, and no problem, I can make changes as you suggest them.

lewismc:
@chrismattmann did you check the license of the plugin code? Unfortunately as mentioned above it is  LGPL licensed. It is fine to have it in the repos, but not OK for us to release the plugin with out shipped releases.

chrismattmann:
crud @lewismc good catch. we should probably at a minimum back this out. I see based on http://cssparser.sourceforge.net/license.html that the license won't work. Sigh. Please feel free to revert. Just trying to push everything forward and missed the license part.

lewismc:
@chrismattmann yeah I was sad about this one as well. I've uploaded the patch (thank you @naegelejd) to the relevant Jira issue for future use and have also reverted the commit. Thanks again @naegelejd for the patch it would have been great to get in to master but we cannot ship the code.

naegelejd:
Of course. I'll keep this in the back of my mind in case I come up with an alternative solution. Thanks guys!

PR #103, opened 2016-04-14
NUTCH-2250 : CommonCrawlDumper : Invalid format and skipped parts

thammegowda:
### Summary
+ Reads all parts of segments
+ writes single document per dump file

### What was the fix?
1. Lists all files insides segments and selects all the parts matching to pattern part-[0-9]{5} under the content directory. 
2. CommonCrawlFormat implementation for "JACKSON" type does not clear state, instead it keeps appending documents. This fix uses a new instance for each document (actually it was hard to resolve the state issue because of too many options such as compressing and archiving).

### DIFF
--- a/src/java/org/apache/nutch/tools/CommonCrawlDataDumper.java
+++ b/src/java/org/apache/nutch/tools/CommonCrawlDataDumper.java
@@ -23,7 +23,6 @@ import java.io.BufferedOutputStream;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
-import java.io.FileFilter;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.MalformedURLException;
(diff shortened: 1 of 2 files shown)


src/java/org/apache/nutch/tools/CommonCrawlDataDumper.java
```diff
@@ -368,7 +346,11 @@ public boolean accept(File file) {
             String mimeType = new Tika().detect(content.getContent());
             // Maps file to JSON-based structure
 
-            jsonData = format.getJsonData(url, content, metadata);
+            //TODO: Make this Jackson Format implementation reusable
```
lewismc: Can you log a Jira issue for this please?

lewismc:
Nice patch. I tried it out with some test segment and have reviewed the code. +1 from me. I'll commit shortly unless objections.

chrismattmann:
+1 looks good to me. commit! :) 👍

src/java/org/apache/nutch/tools/CommonCrawlDataDumper.java
thammegowda: Yes, [NUTCH-2251](https://issues.apache.org/jira/browse/NUTCH-2251)

PR #112, opened 2016-05-17
NUTCH-2262 Utilize parameterized logging notation across Fetcher

lewismc:
This issue addresses https://issues.apache.org/jira/browse/NUTCH-2262


lewismc:
Any objections to commit?

PR #118, opened 2016-05-25
fix for NUTCH-2234 and NUTCH-2236

naegelejd:
Upgrade Elasticsearch and Lucene dependencies, which, in turn, requires updates to Guava and Hadoop dependencies:

- Elasticsearch 1.4.1 -> Elasticsearch 2.3.3
- Lucene 4.10.2 -> 5.5.0
- Guava 16.0.1 -> Guava 18.0
- Hadoop 2.4.0 -> 2.7.2

### DIFF
--- a/ivy/ivy.xml
+++ b/ivy/ivy.xml
@@ -52,7 +52,7 @@
         <dependency org="com.tdunning" name="t-digest" rev="3.1" />
             
         <!-- Hadoop Dependencies -->
-		<dependency org="org.apache.hadoop" name="hadoop-common" rev="2.4.0" conf="*->default">
+		<dependency org="org.apache.hadoop" name="hadoop-common" rev="2.7.2" conf="*->default">
 			<exclude org="hsqldb" name="hsqldb" />
 			<exclude org="net.sf.kosmosfs" name="kfs" />
 			<exclude org="net.java.dev.jets3t" name="jets3t" />
@@ -60,9 +60,9 @@
 			<exclude org="org.mortbay.jetty" name="jsp-*" />
 			<exclude org="ant" name="ant" />
 		</dependency>
-        <dependency org="org.apache.hadoop" name="hadoop-hdfs" rev="2.4.0" conf="*->default"/>
-        <dependency org="org.apache.hadoop" name="hadoop-mapreduce-client-core" rev="2.4.0" conf="*->default"/>
-        <dependency org="org.apache.hadoop" name="hadoop-mapreduce-client-jobclient" rev="2.4.0" conf="*->default"/>
+        <dependency org="org.apache.hadoop" name="hadoop-hdfs" rev="2.7.2" conf="*->default"/>
+        <dependency org="org.apache.hadoop" name="hadoop-mapreduce-client-core" rev="2.7.2" conf="*->default"/>
+        <dependency org="org.apache.hadoop" name="hadoop-mapreduce-client-jobclient" rev="2.7.2" conf="*->default"/>
         <!-- End of Hadoop Dependencies -->
 
 		<dependency org="org.apache.tika" name="tika-core" rev="1.12" />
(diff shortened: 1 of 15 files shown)


ivy/ivy.xml
```diff
@@ -105,6 +105,10 @@
 		<dependency org="org.mortbay.jetty" name="jetty-client" rev="6.1.22" conf="test->default" />
 		<dependency org="org.mortbay.jetty" name="jetty" rev="6.1.22" conf="test->default" />
 		<dependency org="org.mortbay.jetty" name="jetty-util" rev="6.1.22" conf="test->default" />
+		<dependency org="tomcat" name="jasper-runtime" rev="5.5.23" conf="test->default" />
```
lewismc: Why are these Tomcat dependencies added?

lewismc:
This looks dynamite.
Can you also update the relevant properties within default.properties e.g. the Javadoc links for Lucene, Solr, etc? An example of this can be seen within the 2.x default.properties.

Thank you very much for this patch @naegelejd

ivy/ivy.xml
naegelejd: The tomcat dependencies were previously pulled in by Hadoop 2.4.0. They are needed for the protocol-http[client] JUnit tests using JSP.

naegelejd:
I added the relevant properties to `default.properties` and included corresponding `<link>s` for the JavaDoc in `build.xml`. I also updated the Solrj dependency from 5.4.1 -> 5.5.0 to match the Lucene version change.

Since this changes the version of a handful of dependencies, should it correspond to a separate JIRA issue, or is it OK that it's attached to NUTCH-2234? I'm thinking, for example, when someone adds this fix to `CHANGES.txt`, the ticket description doesn't match the actual fix (i.e. this upgrades Elasticsearch to 2.3.3, not 2.1.1, etc.).

lewismc:
Hi @naegelejd dynamite
> I added the relevant properties to default.properties and included corresponding <link>s for the 
> JavaDoc in build.xml. I also updated the Solrj dependency from 5.4.1 -> 5.5.0 to match the Lucene
> version change. 

Excellent thank you

> Since this changes the version of a handful of dependencies, should it correspond to a separate JIRA issue, 

I don't think so no. I think it it best if we adopt a policy that these are kept in sync (to the best of our ability) from now on. This has been an issue previously with dependencies like httpclient, etc. however from now on dependency management should be better handled. I think what you are doing is the right way to go.

> or is it OK that it's attached to NUTCH-2234? I'm thinking, for example, when someone addsthis fix to CHANGES.txt, the ticket description doesn't match the actual fix (i.e. this upgrades Elasticsearch to 2.3.3, not 2.1.1, etc.).

we don't need to add anything to changes.txt apart from the Jira release report. The release announcement is a different kettle of fish and should be more user friendly. 

One thing on the side here. Every pull request and commit message should reflect the Jira tickey title. This way all of our Github hooks work flawlessly. This way it also makes it much easier to track more complex issues like this one where possibly more than one feature is being touched at any one time.

lewismc:
@sujen1412 can you please review proposed changes to scoring-similarity and comment if appropriate?

lewismc:
If @sujen1412 review's changes to scoring-similarity and provides positive feedback then I am +1 on this for merge into master. Thank you @naegelejd

sujen1412:
Looks good to me. Thank you @naegelejd and @lewismc

naegelejd:
Would it make sense to remove the Lucene dependency from `ivy/ivy.xml` here and now, since this PR also updates other dependencies in `ivy/ivy.xml` as well as a few plugins? I've done this locally and all tests pass.

lewismc:
Yes, if you can update the PR then that would be grand.

On Mon, Jun 27, 2016 at 1:45 PM, Joseph Naegele <notifications@github.com>
wrote:

> Would it make sense to remove the Lucene dependency from ivy/ivy.xml here
> and now, since this PR also updates other dependencies in ivy/ivy.xml as
> well as a few plugins? I've done this locally and all tests pass.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/apache/nutch/pull/118#issuecomment-228869736>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABHJlxR22ZrE_8IrGVRwmazFHihHDBYsks5qQDZQgaJpZM4Im2Ll>
> .
>



-- 
*Lewis*

naegelejd:
Good to go.

PR #121, opened 2016-06-19
NUTCH-2283 Fix bad substitution bash error for cassandra docker scripts

lancergr:
"Bad substitution" error appears when running the following docker scripts:
    docker/cassandra/bin/start.sh
    docker/cassandra/bin/stop.sh
    docker/cassandra/bin/restart.sh

e.g.
./bin/start.sh: 18: ./bin/start.sh: Bad substitution


lewismc:
Thank you @lancergr nice :)

PR #129, opened 2016-06-27
NUTCH-2267 - Solr and Hadoop JAR mismatch

sjwoodard:
Explicitly pass in an instance of SystemDefaultHttpClient to CloudSolrClient, otherwise SolrJ will use a default implementation of CloseableHttpClient, which is not present in the HttpClient and HttpCore JARs in Hadoop < 2.8 (see https://issues.apache.org/jira/browse/SOLR-7948 and https://issues.apache.org/jira/browse/HADOOP-12767).

### DIFF
--- a/src/plugin/indexer-solr/src/java/org/apache/nutch/indexwriter/solr/SolrUtils.java
+++ b/src/plugin/indexer-solr/src/java/org/apache/nutch/indexwriter/solr/SolrUtils.java
@@ -22,6 +22,7 @@ import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.mapred.JobConf;
+import org.apache.http.impl.client.SystemDefaultHttpClient;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.impl.HttpSolrClient;
 import org.apache.solr.client.solrj.impl.CloudSolrClient;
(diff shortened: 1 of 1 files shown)


lewismc:
I am +1 on this and will merge shortly unless objections

lewismc:
@sjwoodard actually, I've reverted this commit. In httpclient 4.4.1 the SystemDefaultHttpClient is deprecated
http://hc.apache.org/httpcomponents-client-4.4.x/httpclient/apidocs/index.html?org/apache/http/impl/client/SystemDefaultHttpClient.html
You are advised to use the HttpClientBuilder instead. 
Can you please update the patch accordingly? Thanks.
http://hc.apache.org/httpcomponents-client-4.4.x/httpclient/apidocs/index.html?org/apache/http/impl/client/HttpClientBuilder.html

sjwoodard:
Sorry, I didn't notice it was deprecated. Using HttpClientBuilder doesn't work because it returns a CloseableHttpClient, which was introduced in HttpClient 4.3 (same version they deprecated SystemDefaultHttpClient). Hadoop is still using HttpClient 4.2.5, which doesn't have CloseableHttpClient. 

So, I think using SystemDefaultHttpClient is the cleanest way to bridge the two HttpClient versions. Otherwise, Hadoop 2.8 will fix this.

As an alternative, users with this issue can upgrade use these Hadoop 2.8 patches to upgrade their HttpClient:

- https://issues.apache.org/jira/browse/HADOOP-12767
- https://issues.apache.org/jira/browse/HADOOP-11859

I'll leave this branch up as a guide for other users with this issue.

PR #134, opened 2016-07-12
NUTCH-2280 fix the cookie policy issue in post form authentication

stevegy:
Fix the [NUTCH-2280](https://issues.apache.org/jira/browse/NUTCH-2280).
When a HTTP post receives session cookie from server, it might hit an issue when the server has a domain name set as ".domain.com". The "." at beginning will cause HttpClient reject this cookie.

### DIFF
--- a/src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpFormAuthentication.java
+++ b/src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpFormAuthentication.java
@@ -31,9 +31,12 @@ import java.util.Set;
 import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.NameValuePair;
+import org.apache.commons.httpclient.cookie.CookiePolicy;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.apache.commons.httpclient.methods.PostMethod;
+import org.apache.commons.httpclient.params.HttpMethodParams;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.reflect.FieldUtils;
 import org.jsoup.Jsoup;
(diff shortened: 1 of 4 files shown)


src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpFormAuthentication.java
```diff
@@ -135,6 +143,26 @@ public boolean getFollowRedirects() {
...
+   */
+  private void setCookieParams(HttpFormAuthConfigurer formConfigurer,
+		  HttpMethodParams params)
+  		throws NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException {
+  	// NUTCH-2280 - set the HttpClient cookie policy
+        if (formConfigurer.getCookiePolicy() != null) {
+      	  String policy = formConfigurer.getCookiePolicy();
+      	  Object p = FieldUtils.readDeclaredStaticField(CookiePolicy.class, policy);
+      	  if(null != p) {
+      		  LOGGER.debug("reflection of cookie value: " + p.toString());
```
lewismc: Code formatting. 2 space indents please.

src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpFormAuthentication.java
```diff
@@ -135,6 +143,26 @@ public boolean getFollowRedirects() {
       }
     }
   }
+  
+  /**
+   * @throws NoSuchFieldException
```
lewismc: Substantiated Javadoc on all of the above would be preferable.

stevegy: Sure. I have updated this source and re-pushed.
And, add an updated httpclient-auth.xml.template for how to configure this cookie policy.

PR #135, opened 2016-07-13
NUTCH-2285 Digest Authentication support for Nutch 2.X REST API.

kamaci:


kamaci:
@lewismc could you check this?

PR #147, opened 2016-08-26
NUTCH-2308 SSL connection test at TestNutchAPI is implemented.

kamaci:


src/test/org/apache/nutch/api/AbstractNutchAPITestBase.java
```diff
@@ -160,6 +177,37 @@ private void initializeSSLProperties() {
     conf.set("restapi.auth.ssl.keypass", "password");
   }
 
+  private void loadCertificate() throws CertificateException, NoSuchAlgorithmException, KeyStoreException, IOException {
+    KeyStore ks = KeyStore.getInstance(KeyStore.getDefaultType());
+    char[] password = "testpassword".toCharArray();
+    String testTrustKeyStorePath = "src/test/testTrustKeyStore";
```
lewismc: This will break on Windows for sure. Please use the classloader.

src/test/org/apache/nutch/api/AbstractNutchAPITestBase.java
```diff
@@ -160,6 +177,37 @@ private void initializeSSLProperties() {
...
+    InputStream localCertIn = new FileInputStream(file);
+
+    ks.load(localCertIn, password);
+    localCertIn.close();
+
+    if (ks.containsAlias("nutch")) {
+      return;
+    }
+
+    InputStream certIn = new FileInputStream("src/test/nutch.cer");
```
lewismc: Please use the classloader.

kamaci: Do you want me to use sth. like this.getClass().getClassLoader().getResource("nutch.cer").getFile() ?

src/test/org/apache/nutch/api/AbstractNutchAPITestBase.java
kamaci: Does this break Windows? I just give the path and I think it won't be a problem. Do I miss anything?

src/test/org/apache/nutch/api/AbstractNutchAPITestBase.java
lewismc: Yes. Whenever we load resources.... it should most likely always be done this way.

src/test/org/apache/nutch/api/AbstractNutchAPITestBase.java
lewismc: Please use classloader.

lewismc:
Tests failing please see Jira.

kamaci:
@lewismc I've updated the PR. Could you apply it again?

PR #150, opened 2016-09-02
NUTCH-2314 indexer-elastic2 plugin is used for javadoc and eclipse targets

kamaci:


lewismc:
Can you also add it to https://github.com/apache/nutch/blob/2.x/default.properties#L160
Thanks

kamaci:
@lewismc I've updated the PR.

PR #152, opened 2016-09-24
Fix for NUTCH-2317: Plugin jars don't get added to classpath

sujen1412:


lewismc:
@sujen please reference the Jira issue in your comment on the PR. It makes it much easier for those of us who review patches. Good job on this one this has been bugging us for a while. Thanks Sujen.

lewismc:
Forgot to say... +1

sujen1412:
Thanks, here is the link to JIRS - https://issues.apache.org/jira/browse/NUTCH-2317

sebastian-nagel:
-1
- this would mean to have all libs from all plugins on the classpath. In consequence, we would have to make sure that there are absolutely no conflicts between all dependencies.
- this solution would have only effect in local mode but not for distributed mode where plugins and their dependencies are in the job file

lewismc:
Excellent points Seb. Some suggestions below

On Sep 25, 2016 2:53 AM, "Sebastian Nagel" <notifications@github.com> wrote:
>
> -1
>
> this would mean to have all libs from all plugins on the classpath.

Your right this is not good as several plugins may not be being used in The
deployment. Maybe we should look into removing the unused plugins from the
copy command when we copy from /build to /runtime. That would solve the
issue you've pointed out. Wdyt?

In consequence, we would have to make sure that there are absolutely no
conflicts between all dependencies.

Is this a bad thing? Imho it is a good thing. It is possibly a naive
position to assume that we can obtain zero conflicts.

> this solution would have only effect in local mode but not for
distributed mode where plugins and their dependencies are in the job file
Correct. That is another issue which has not been addressed yet.

sebastian-nagel:
The advantage of the current implementation and the plugin class loader: two independent plugins can have conflicting library versions as dependencies. There may be still conflicts between "base" libs and plugin dependencies, see https://issues.apache.org/jira/browse/NUTCH-2316 for such an issue.

To add only libs from activated plugins to the class path would only make conflicts less probable. But we allow users to use any combination of available plugins, so sooner or later we see a conflict as a Jira issue. Also making the build use a runtime configuration such as the property `plugin.includes` would violate the idea that users always can deploy their configuration by just placing the customized nutch-site.xml in a folder which comes early on the classpath. 

> In consequence, we would have to make sure that there are absolutely no conflicts between all dependencies. Is this a bad thing? Imho it is a good thing. It is possibly a naive position to assume that we can obtain zero conflicts.

Yes, of course, it would be the best solution if we could resolve all conflicts. No question. But there are many conflicts, e.g., currently we have in runtime/local:
```
lib/httpclient-4.3.5.jar
plugins/lib-htmlunit/httpclient-4.3.4.jar
plugins/lib-selenium/httpclient-4.5.1.jar
plugins/indexer-solr/httpclient-4.4.1.jar
plugins/indexer-cloudsearch/httpclient-4.3.6.jar
```
And the right way would be: 1. resolve the conflicts and make sure that everything runs fine, 2. configure one global classpath. To get NUTCH-2316 resolved could be the first step of the first task.

HiranChaudhuri:
The need for adding these jars may be similar to what I tried to overcome with NUTCH-2429.

sebastian-nagel:
Closing, see [NUTCH-2317](https://issues.apache.org/jira/browse/NUTCH-2317).

PR #153, opened 2016-10-06
Bug in setting default linkdb path

sachin086:
There was an extra if condition in run( ) method where it sets linkdb path. It is evident when we use nutch REST api.

https://github.com/apache/nutch/blob/master/src/java/org/apache/nutch/indexer/IndexingJob.java#L272

### DIFF
--- a/src/java/org/apache/nutch/indexer/IndexingJob.java
+++ b/src/java/org/apache/nutch/indexer/IndexingJob.java
@@ -269,7 +269,6 @@ public class IndexingJob extends NutchTool implements Tool {
     List<Path> segments = new ArrayList<Path>();
 
     if(args.containsKey(Nutch.ARG_LINKDB)){
-      if(args.containsKey(Nutch.ARG_LINKDB)) {
         Object path = args.get(Nutch.ARG_LINKDB);
         if(path instanceof Path) {
           linkdb = (Path) path;
@@ -277,8 +276,7 @@ public class IndexingJob extends NutchTool implements Tool {
         else {
           linkdb = new Path(path.toString());
         }
-      }
-      else {
+    } else {
         linkdb = new Path(crawlId+"/linkdb");
       }
     }
--- a/src/java/org/apache/nutch/indexer/IndexingJob.java
+++ b/src/java/org/apache/nutch/indexer/IndexingJob.java
@@ -279,7 +279,6 @@ public class IndexingJob extends NutchTool implements Tool {
     } else {
         linkdb = new Path(crawlId+"/linkdb");
       }
-    }
 
     if(args.containsKey(Nutch.ARG_SEGMENTDIR)){
       isSegment = true;


sebastian-nagel:
Good catch! Would you mind also to open an issue report at https://issues.apache.org/jira/browse/NUTCH, to get this properly tracked and added to the release notes. Thanks.

Since the LinkDb is optional and not required to run the IndexingJob, needs to be discussed whether the intention wasn't originally
```
    if(args.containsKey(Nutch.ARG_LINKDB)){
      if(args.get(Nutch.ARG_LINKDB) != null) {
```
But add an empty directory as linkdb should not harm in any case. Your fix should work, thanks!

sachin086:
Hi, I have created an issue. You can find it at : https://issues.apache.org/jira/browse/NUTCH-2324

sujen1412:
Nice find, thank you @sachin086!

@sebastian-nagel you're right, the intention was to have it like 
```   
 if(args.containsKey(Nutch.ARG_LINKDB)){
    if(args.get(Nutch.ARG_LINKDB) != null) {
```

This was because the other commands in the REST API do not require the user to give the full path of the crawldb, segment, linkdb directories as they are all mapped through their corresponding crawlIDs. 
Hence, the args {} containing and key=linkdb with value=null would act as a flag indicating that the user wants to use the linkdb associated with this crawlId for indexing. 
Or else, if the value=<some path> then that would be used for the linkdb. 

If there is a better way to achieve this please suggest, my thought to do it this way was to keep the format of requests for the job endpoint consistent across different jobs.

sebastian-nagel:
Thanks, @sujen1412!  Probably, the index job was the first job which takes an optional argument. How the 3 possible values (custom path, default path, no linkdb) are encoded does not really matter. The way should be consistent. Currently we have:

1. key exists in args and value != null :: custom path
2. key does not exist :: default path (crawlId/linkdb or crawlId/crawldb etc.)

If we keep this unchanged and continue

3. key exists and value == null :: do not use

Btw., looks like linkdb is only half integrated: o.a.n.service.JobManager.JobType contains INVERTLINKS while o.a.n.webui.client.model.JobInfo.JobType doesn't?

sujen1412:
Thank you @sebastian-nagel. Created an issue for implementing invertlinks in webui - https://issues.apache.org/jira/browse/NUTCH-2326. 

What do you think about having it in this manner : 
1. key exists and value != null :: use custom path 
2. Key exists and value == null :: use default path ($crawlId/linkdb)
3. key does not exist :: do not use linkdb.

sebastian-nagel:
Hi,

> NUTCH-2326

Great.

>  3. key does not exist :: do not use linkdb. 

The only point I see: for other tools 3 indicates that the default path is used, e.g. CrawlDb:
```
if (args.containsKey(Nutch.ARG_CRAWLDB)) {
  ...
} else {
  crawlDb = new Path(crawlId+"/crawldb");
}
```

But that's a weak argument, and there is no reason why not to have a slightly different logic for optional arguments. So, +1: go ahead!

sujen1412:
Thanks @sachin086 and @sebastian-nagel. 

I will update the PR as per the conversation on this thread.

lewismc:
@sujen1412 going to merge this into master. Please open a new ticket for the proposed additional fixes. Thanks @sachin086

PR #155, opened 2016-10-20
Fix for NUTCH-2327: Seeds injected in REST must be ingested into HDFS

sujen1412:


lewismc:
tested with master branch. +1

PR #156, opened 2016-11-02
fix for NUTCH-2269 contributed by r0ann3l

r0ann3l:
I found when you try to commit the documents, the writers are closed. This is because the close method is invoked first and then the commit method. Then, the following exception is thrown: java.lang.IllegalStateException: Connection pool shut down


marconett:
I applied this patch and now the clean job runs without error.

Should the "Num Docs" count in my core job change after a successful clean?
The clean job you returned "SolrIndexer: deleting 28/28 documents", but the "Num Docs" count didn't change.

PR #161, opened 2016-12-13
Fix for NUTCH-2046 contributed by jnioche

jnioche:
This makes the seed argument optional and is an alternative to the solution proposed in [https://issues.apache.org/jira/browse/NUTCH-2046]. The latter is also acceptable and has the advantage of not breaking compatibility.

### DIFF
--- a/src/bin/crawl
+++ b/src/bin/crawl
@@ -15,14 +15,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-# Usage: crawl [-i|--index] [-D "key=value"] [-w|--wait] <Seed Dir> <Crawl Dir> <Num Rounds>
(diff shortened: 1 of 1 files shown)


lewismc:
+1 for me @jnioche 
@sebastian-nagel you had made a comment over on NUTCH-2046, are you happy with this approach? I've tested in locally and like it.

sebastian-nagel:
+1 from my side, but we could add a note to CHANGES.txt (as part of a section about API-breaking changes). Users need to update scripts calling bin/crawl.

lewismc:
@jnioche

jnioche:
thanks for the reviews and the nudge. Will add a comment on CHANGES in a separate commit.

PR #166, opened 2017-01-09
NUTCH-2347 Logger is used instead of printing Throwable.

kamaci:


src/java/org/apache/nutch/parse/ParseUtil.java
```diff
@@ -237,7 +237,7 @@ public void processSitemapParse(String url, WebPage page,
           try {
             reversedUrl = TableUtil.reverseUrl(toUrl); // collect it
           } catch (MalformedURLException e) {
-            e.printStackTrace();
+            LOG.error("MalformedURLException occurred: ", e);
```
sebastian-nagel: That a specific exception happened is obvious, would be better to explain what happened with which input, e.g.
```
LOG.error("Failed to reverse URL {}: {}", toUrl, e.getMessage());
```
This also applies to many other of the new log messages.

sebastian-nagel: Hi @kamaci, thanks. That's of course better than the current state. But at least some messages could be more informative.

lewismc:
@sebastian-nagel are you suggesting that each message is customized?

sebastian-nagel:
Explicitly logging the exception class is redundant, e.g.:
```
2017-01-18 18:25:02,575 ERROR parse.ParseUtil - MalformedURLException occurred: 
java.net.MalformedURLException: no protocol: ://example.com/
        at java.net.URL.<init>(URL.java:593)
        ...
```
Without a meaningful message we could also log:
```
LOG.error(StringUtils.stringifyException(e));
```
A message which explaines what went wrong would be better. But as I said, it's a further improvement. :)

kamaci:
@sebastian-nagel @lewismc  I've created another PR to get a better logging. I'll close this one.

PR #167, opened 2017-01-09
NUTCH-2348 GZIPInputStream is closed at finally.

kamaci:


sebastian-nagel:
Hi @kamaci, have you tested whether this still allows to iterate over all records of an [ARC file](http://crawler.archive.org/articles/developer_manual/arcs.html)? I suspect that GZipInputStream.close() will implicitly close the underlying FSDataInputStream which would cause that any subsequent records are skipped. Usually every ARC record is compressed on its own, that's why a new GZipInputStream is used per record.

kamaci:
@sebastian-nagel I've closed the input stream at finally block. It causes a memory leak when you don't close such input streams. unzipBestEffort method at GZIPUtils.java also closes the input stream at such case. 

I can add a test method test to explicitly test it (or do it manually) whether it causes an unintended case or not if you can guide me about how we can understand it?

sebastian-nagel:
@kamaci, after a look at the Java sources: calling [GZIPInputStream.close()](http://code.metager.de/source/xref/openjdk/jdk8/jdk/src/share/classes/java/util/zip/GZIPInputStream.java#129) calls [InflaterInputStream.close()](http://code.metager.de/source/xref/openjdk/jdk8/jdk/src/share/classes/java/util/zip/InflaterInputStream.java#218) which closes the underlying stream. In ArcRecordReader that's `FSDataInputStream in` which should be kept open until `ArcRecordReader.close()` is called, but not every time `next(...)` is executed (with and without an exception). Only if `ArcRecordReader.close()` is never called the input stream may leak as resource (usually the max. number of open files per process is limited).

GZIPUtils.unzipBestEffort() is different because the underlying ByteArrayInputStream is hold by a local variable.

Given that ArcRecordReader is rarely used (for 1.x it can be used to convert content from ARC files into a Nutch segment, cf. ArcSegmentCreator), I wouldn't personally spent much time for investigation, although I'm curious which opinion is correct. For 2.x we also could consider to remove all ARC related classes.

kamaci:
@sebastian-nagel OK, then I'm closing the PR.

PR #171, opened 2017-01-17
NUTCH-2351 Logging with generic class name.

kamaci:


sebastian-nagel:
Thanks for the careful analysis and the other unifications made: always use LOG, replace commons logging by slf4j.

PR #175, opened 2017-01-21
Fix for NUTCH-2353 contributed by jorgelbg

jorgelbg:
This PR adds the possibility of attaching metadata to the seed file created using the REST API as showed on https://issues.apache.org/jira/browse/NUTCH-2353, the supported syntax is:

```json
{
    "name":"name-of-seedlist", 
    "seedUrls":[
        {
            "url" : "http://example.com",
            "metadata" : {
                "key1" : "value1",
                "key2" : "value2",
                "key3" : "value3"
            }
        }
    ]
}
```

Also the old syntax is supported:

```json
{
    "name":"name-of-seedlist", 
    "seedUrls":["http://www.example.com", "http://www.example1.com"]
}
```

### DIFF
--- a/src/java/org/apache/nutch/service/model/request/SeedUrl.java
+++ b/src/java/org/apache/nutch/service/model/request/SeedUrl.java
@@ -17,6 +17,8 @@
 package org.apache.nutch.service.model.request;
 
 import java.io.Serializable;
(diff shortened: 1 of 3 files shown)


lewismc:
Nice @jorgelbg +1

PR #177, opened 2017-02-09
NUTCH-2357 Index metadata throw Exception because writable object can…

sebastian-nagel:
…not be cast to Text

- do not cast CrawlDatum metadata values from o.a.hadoop.io.Writable to o.a.hadoop.io.Text
(contributed by Eyeris Rodriguez Rueda)


chrismattmann:
LGTM!

PR #178, opened 2017-02-13
NUTCH-2359 RegexParseFilter: ill-formed rules raise error

laknath:
- README
  - Using any whitespace character(\s) instead of tab(\t) as rule delimiter

Issue: https://issues.apache.org/jira/browse/NUTCH-2359


sebastian-nagel:
Committed to 1.x/master (9a9c4b3).

PR #181, opened 2017-04-04
fix fetcher.throughput.threshold.check.after

knap1930:
Without adding on System.currentTimeMillis() this can't possibly work?

According to nutch-default.xml the default value should be 5 minutes.

### DIFF
--- a/src/java/org/apache/nutch/fetcher/FetcherReducer.java
+++ b/src/java/org/apache/nutch/fetcher/FetcherReducer.java
@@ -861,8 +861,8 @@ public class FetcherReducer extends
       LOG.info("Fetcher: throughput threshold sequence: "
           + throughputThresholdSequence);
     }
-    long throughputThresholdTimeLimit = conf.getLong(
-        "fetcher.throughput.threshold.check.after", -1);
+    long throughputThresholdTimeLimit = System.currentTimeMillis() + (conf.getLong(
+        "fetcher.throughput.threshold.check.after", 5) * 60 * 1000);
 
     do { // wait for threads to exit
       pagesLastSec = pages.get();


sebastian-nagel:
Good catch, thanks! Would you mind open also an issue on the [Apache Nutch issue tracker](https://issues.apache.org/jira/browse/NUTCH)? Just to make sure the fix is tracked and added to the release notes.

src/java/org/apache/nutch/fetcher/FetcherReducer.java
```diff
@@ -861,8 +861,8 @@ public void run(Context context) throws IOException, InterruptedException {
       LOG.info("Fetcher: throughput threshold sequence: "
           + throughputThresholdSequence);
     }
-    long throughputThresholdTimeLimit = conf.getLong(
-        "fetcher.throughput.threshold.check.after", -1);
```
sebastian-nagel: This would set the time when the throughput threshold is check per task not per job. It makes a difference if a task is started later or is restarted. In analogy to "fetcher.timelimit.mins" and how it's done in 1.x, I would set it (overwrite it with the effective system time) in `FetcherJob.run(Map<String, Object> args)`.

sjwoodard:
It looks like the same problem exists in the v1 Fetecher, correct?

sebastian-nagel:
Why Nutch 1.x should be affected? Could you explain "looks like"?

sjwoodard:
I was working on a related issue and noticed the code looked similar. I just got around to testing 1.x, and it's fine. Thanks!

src/java/org/apache/nutch/fetcher/FetcherReducer.java
sebastian-nagel: Sorry, after thinking again about it - you're right. Fetcher differs between 1.x and 2.x in one important point:
- (1.x) fetching is taking place in the map phase
- (2.x) pages are fetched in the reduce phase.
Ok, the property `fetcher.timelimit.mins` is set absolutely for the whole job. For scheduling cycles it's important that the end time is per job and not per task. For 2.x users must include the map phase. The throughput threshold check is not really about limiting the time spent for one crawl/fetch cycle but to avoid tasks consuming resources if there are only few or slow queues. It's definitely ok to set this limit relative per reduce task.

sebastian-nagel:
Thanks, @knap1930!

PR #187, opened 2017-04-20
NUTCH-2376 Improve configurability of HTTP Accept* header fields

sebastian-nagel:
- Accept, Accept-Language and Accept-Charset are configured and used the
  same way for both protocol-http and protocol-httpclient
- a space as value will unset these header fields (not sent in request)
  I've tried to overwrite the default by an empty value but that's not possible with the Hadoop configuration mechanism. Of course, one could configure instead `*` as "Accept-Language" or "Accept-Charset" (resp. `*/*` as "Accept" content type) but that will unnecessarily blow up the HTTP requests.


lewismc:
+1

PR #191, opened 2017-05-23
NUTCH-2388 bin/crawl indexing only webpages of current batch instead of all

kaidul:
During each iteration, after generating, fetching, parsing and updating the current batch into DB, the indexer is supposed to index the current batch too. But its indexing all currently.
```shell
__bin_nutch index $commonOptions -D solr.server.url=$SOLRURL -all -crawlId "$CRAWL_ID"
```

It should be like below i guess -
```shell
__bin_nutch index $commonOptions -D solr.server.url=$SOLRURL $batchId -crawlId "$CRAWL_ID"
```


lewismc:
Thanks @kaidul

PR #196, opened 2017-07-04
Nutch 2397

sebastian-nagel:
(improved solution contributed by Vipul Behl):
 - do not add superfluous line breaks and space in parsed text
 - fix units tests
 - also fix parse-html in addition to parse-tika


sebastian-nagel:
Merged into master 48c38b0.

PR #197, opened 2017-07-04
NUTCH-2395 Cannot run job worker! - error while running multiple crawling jobs in parallel

lewismc:
This PR addresses https://issues.apache.org/jira/browse/NUTCH-2395


src/java/org/apache/nutch/crawl/GeneratorJob.java
```diff
@@ -153,6 +153,11 @@ public void set(String url, float score) {
     public SelectorEntryComparator() {
       super(SelectorEntry.class, true);
     }
+
+    @Override
+    synchronized public int compare(byte[] b1, int s1, int l1, byte[] b2, int s2, int l2) {
```
sebastian-nagel: Making a method synchronized penalizes every call, even in the default case: one single reducer running per JVM. Shouldn't the right solution be to implement a thread-safe comparator? Implementing a thread-safe version of compare should be everything to do:
```
    @Override
    public int compare(byte[] b1, int s1, int l1, byte[] b2, int s2, int l2) {
      SelectorEntry key1 = new SelectorEntry();
      SelectorEntry key2 = new SelectorEntry();
      DataInputBuffer buffer = new DataInputBuffer();

      try {
        buffer.reset(b1, s1, l1);
        key1.readFields(buffer);
        
        buffer.reset(b2, s2, l2);
        key2.readFields(buffer);
        
      } catch (IOException e) {
        throw new RuntimeException(e);
      }
      
      return compare(key1, key2);
    }
```

sebastian-nagel:
Closing unmerged PR for 2.x (NUTCH-2395 also has been closed). Thanks, @lewismc!

PR #200, opened 2017-07-06
NUTCH-2399: Support multivalue fields.

pipldev:
Currently, if there is a NutchField with multiple values, only the first value is indexed (because this is what doc.getFieldValue returns). This fix checks if the NutchField has multiple values, they are added as an array (multivalue) field.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1041,6 +1041,7 @@
         <source path="${plugins.dir}/indexer-elastic/src/java/" />
         <source path="${plugins.dir}/indexer-elastic/src/test/" />
         <source path="${plugins.dir}/indexer-elastic-rest/src/java/"/>
+        <source path="${plugins.dir}/index-jexl-filter/src/java/" />
         <source path="${plugins.dir}/index-metadata/src/java/" />
         <source path="${plugins.dir}/index-more/src/java/" />
         <source path="${plugins.dir}/index-more/src/test/" />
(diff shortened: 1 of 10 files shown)


lewismc:
Hi @pipldev thank you for the patch... can you please open a JIRA issue over at https://issues.apache.org/jira/projects/NUTCH
Additionally, this change would need to be implemented for all indexers not just the elasticsearch indexer.

sebastian-nagel:
indexer-solr already indexes all values. That's why it's mandatory to declare such fields (eg. anchors) in the [schema.xml](https://github.com/apache/nutch/blob/master/conf/schema.xml#L385) as `multiValued="true"`. But yes, it's a good idea to check the remaining indexers as well.

pipldev:
Opened a JIRA issue: https://issues.apache.org/jira/browse/NUTCH-2399.
I don't think I can test the other indexers, as I am not familiar with them and don't have them installed, but seeing as indexer-solr already implements it, it seems like this patch can go forward independently.

thilohaas:
Also submitted a patch for `indexer-elastic-rest` https://github.com/apache/nutch/pull/203

jorgelbg:
This PR contains also the commits for #219 (NUTCH 2415) going to extract the patch to `indexer-elastic` into a separated PR and close this one.

pipldev:
Oops, sorry for that. Not sure how it happened.
Only the a614dd4 commit should be here.

jorgelbg:
@pipldev Don't worry, I'll just open a new PR with your changes. Thanks for the contribution, this was a good catch!

PR #206, opened 2017-07-21
NUTCH-2403: Fix spelling of phantomJS configuration

mfeltscher:
The Nutch Selenium documentation states that PhantomJS can be used as `phantomJS` for `selenium.driver`. The correct value would be `phantomjs` according to https://github.com/apache/nutch/blob/master/src/plugin/lib-selenium/src/java/org/apache/nutch/protocol/selenium/HttpWebClient.java#L124


lewismc:
Thank you @mfeltscher

PR #211, opened 2017-08-12
NUTCH-1932 Automatically remove orphaned pages

sebastian-nagel:
- apply Markus Jelsma's latest patch, 2016-06-30
- add method orphanedScore(Text, CrawlDatum) to ScoringFilter interface
- complete unit tests for CrawlDb update

### DIFF
--- a/src/java/org/apache/nutch/scoring/ScoringFilter.java
+++ b/src/java/org/apache/nutch/scoring/ScoringFilter.java
@@ -164,7 +164,7 @@ public interface ScoringFilter extends Configurable, Pluggable {
    *          newly discovered page. If not null, filters should use score
    *          values from this parameter as the starting values - the
    *          <code>datum</code> parameter may contain values that are no longer
-   *          valid, if other updates occured between generation and this
+   *          valid, if other updates occurred between generation and this
    *          update.
    * @param datum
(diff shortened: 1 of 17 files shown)


src/java/org/apache/nutch/scoring/ScoringFilter.java
```diff
@@ -179,6 +179,20 @@ public void updateDbScore(Text url, CrawlDatum old, CrawlDatum datum,
...
 
   /**
+   * This method may change the score or status of CrawlDatum during CrawlDb
+   * update, when the URL is neither fetched nor has any inlinks.
+   *
+   * @param url
+   *          URL of the page
+   * @param datum
+   *          CrawlDatum for page
+   * @throws ScoringFilterException
```
lewismc: I think this 'may' break Javadoc generation if no comment is provided alongside the Exception itself.

sebastian-nagel: Ok, thanks. However, the current state would require more fixes. `@throws ScoringFilterException` hasn't a description anywhere, it's even not consistently "declared" in the Javadoc. Let's keep this for a later improvement.

sebastian-nagel:
Merged after rebase and squash.

PR #212, opened 2017-08-12
NUTCH-2408 CrawlDb: allow update from unparsed segments

sebastian-nagel:
- use unparsed segments to update status in CrawlDb
- but log if unparsed segment is detected (priorly unparsed segments where logged and skipped)


lewismc:
+1

PR #217, opened 2017-08-28
NUTCH-2414 - Allow LanguageIndexingFilter to actually filter documents by language

pipldev:
Added property lang.index.languages. If exists and is not empty, it is treated as a comma-separated list of languages to index. A document in another language will not be indexed.
"unknown" is a valid language code.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -1566,6 +1566,16 @@ visit https://wiki.apache.org/nutch/SimilarityScoringFilter-->
   </description>
 </property>
 
+<property>
+  <name>lang.index.languages</name>
+  <value></value>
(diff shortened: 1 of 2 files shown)


lewismc:
[Markus' comments](https://issues.apache.org/jira/browse/NUTCH-2414?focusedCommentId=16144237&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16144237) over at the JIRA issue are certainly the better solution. This patch does however address the original item description and based on that I feel we should merge into master branch.
We can revisit once a Jexl IndexingFilter is developed.

lewismc:
Thank you @pipldev please feel free to implement the Jexl filter if you are interested.

PR #218, opened 2017-08-28
fix for NUTCH-1480 contributed by r0ann3l

r0ann3l:
With this patch now we can have many instances of the same IndexWriter class, but with different configurations. Also, we can copy, rename or remove fields of documents for every index writer individually. Besides, the parameters needed by the index writers will be into separated XML files, so them will be not into nutch-site.xml anymore.


jorgelbg:
This PR includes more changes than the original ticket and breaks BC with custom indexers. @r0ann3l could you squash all the changes into a single commit? that would help in the review process since this PR has a lot of changes.

odisleysi:
I like this idea. I work in a project that needs to save documents in solr for searching and in elasticsearch for statistics. This solve the problem.

src/java/org/apache/nutch/indexer/IndexWriterConfig.java
```diff
@@ -0,0 +1,95 @@
+package org.apache.nutch.indexer;
```
lewismc: Please add license header

src/java/org/apache/nutch/indexer/IndexWriterConfig.java
```diff
@@ -0,0 +1,95 @@
...
+
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+import java.util.*;
```
lewismc: Please use explicit imports instead of wildcard

src/java/org/apache/nutch/indexer/IndexWriters.java
```diff
@@ -16,132 +16,245 @@
...
-          for (int i = 0; i < extensions.length; i++) {
-            Extension extension = extensions[i];
-            IndexWriter writer = (IndexWriter) extension.getExtensionInstance();
-            LOG.info("Adding " + writer.getClass().getName());
-            if (!indexerMap.containsKey(writer.getClass().getName())) {
-              indexerMap.put(writer.getClass().getName(), writer);
+
+          HashMap<String, Extension> extensionMap = new HashMap<>();
+          for (Extension extension : extensions) {
+            LOG.info("Index writer " + extension.getClazz() + " identified.");
```
lewismc: Please use parameterized logging

src/java/org/apache/nutch/indexer/MappingReader.java
```diff
@@ -0,0 +1,93 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nutch.indexer;
+
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+import java.util.*;
```
lewismc: Do not use wildcard

src/plugin/indexer-rabbit/src/java/org/apache/nutch/indexwriter/rabbit/RabbitMQConstants.java
```diff
@@ -17,28 +17,26 @@
 package org.apache.nutch.indexwriter.rabbit;
 
 interface RabbitMQConstants {
-    String RABBIT_PREFIX = "rabbitmq.indexer";
```
lewismc: Why did you remove all of these?

src/plugin/indexer-solr/src/java/org/apache/nutch/indexwriter/solr/SolrConstants.java
```diff
@@ -19,19 +19,23 @@
 public interface SolrConstants {
   public static final String SOLR_PREFIX = "solr.";
```
lewismc: Any reason to remove all of these as well?

src/plugin/indexer-solr/src/java/org/apache/nutch/indexwriter/solr/SolrIndexWriter.java
```diff
@@ -75,19 +70,61 @@
...
+        }
+        break;
+      case "cloud":
+        for (String url : urls) {
+          CloudSolrClient sc = SolrUtils.getCloudSolrClient(url);
+          sc.setDefaultCollection(parameters.get(SolrConstants.COLLECTION));
+          solrClients.add(sc);
+        }
+        break;
+      case "concurrent":
```
lewismc: Can you throw unsupported Exception at this stage? and also a default case?

src/plugin/indexer-rabbit/src/java/org/apache/nutch/indexwriter/rabbit/RabbitIndexWriter.java
```diff
@@ -31,164 +31,182 @@
...
+
+  public String describe() {
+    return "RabbitIndexWriter\n" +
+            "\t" + RabbitMQConstants.SERVER_HOST + " : Host of RabbitMQ server\n" +
+            "\t" + RabbitMQConstants.SERVER_PORT + " : Port of RabbitMQ server\n" +
+            "\t" + RabbitMQConstants.SERVER_VIRTUAL_HOST + " : Virtualhost name\n" +
+            "\t" + RabbitMQConstants.SERVER_USERNAME + " : Username for authentication\n" +
+            "\t" + RabbitMQConstants.SERVER_PASSWORD + " : Password for authentication\n" +
+            "\t" + RabbitMQConstants.COMMIT_SIZE + " : Buffer size when sending to RabbitMQ (default 250)\n";
+  }
```
jorgelbg: For consistency probably use the `StringBuilder` pattern here?

conf/index-writers.xsd
```diff
@@ -0,0 +1,179 @@
...
+      <xs:documentation>
+        Root tag of index-writers.xml document. It's a wrapper for the all index writers.
+      </xs:documentation>
+    </xs:annotation>
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="writer" type="writerType" maxOccurs="unbounded" minOccurs="1">
+          <xs:annotation>
+            <xs:documentation>
+              Contains the all configuration of a particular index writer.
```
jorgelbg: typo? Also it would be a good idea to have empty lines at the end of this file and the `conf/index-writers.xml.template` for git/diff compatiblity.

src/plugin/indexer-rabbit/src/java/org/apache/nutch/indexwriter/rabbit/RabbitMQConstants.java
r0ann3l: Hi @lewismc. The prefix is not necessary anymore. The new structure allows us to have the same key of a parameter to many index writers without ambiguity or confusion. The prefix makes a parameter key larger and really I do not believe that is necessary.

r0ann3l:
Thanks @lewismc and @jorgelbg for your reviews. All your comments have been fixed.

lewismc:
Tested this on Solr 6 and works well... any comments folks?

sebastian-nagel:
Looks good! I've tried to use indexer-dummy with this PR applied - it took long to configure the index-writers.xml properly, so we should definitely add "stub" sections for all index writers which are (still) based on configuration properties. All index writers should work out-of-the-box!

conf/index-writers.xml.template
```diff
@@ -0,0 +1,75 @@
...
+        <field source="metatag.description" dest="description"/>
+        <field source="metatag.keywords" dest="keywords"/>
+      </rename>
+      <remove>
+        <field source="content"/>
+        <field source="segment"/>
+        <field source="boost"/>
+      </remove>
+    </mapping>
+  </writer>
```
sebastian-nagel: Add stub <writer> sections for all indexer-* plugins so that they work out-of-the-box without modifications of the index-writers.xml required, e.g. for indexer-dummy:
```
  <writer id="indexer_dummy_1" class="org.apache.nutch.indexwriter.dummy.DummyIndexWriter">
    <parameters>
      <param name="dummy-dummy" value="dummy-dummy"/>
    </parameters>
    <mapping>
      <copy>
        <field source="dummy-dummy" dest="dummy-dummy"/>
      </copy>
      <rename/>
      <remove/>
    </mapping>
  </writer>
```
That's long for a dummy section, but the schema (index-writers.xsd) and the IndexWriters class requires all the elements and attributes. Maybe it's better to "relax" the schema, make elements/attributes optional and make IndexWriters not fail with NPEs.

r0ann3l:
Thanks @sebastian-nagel  for your review. Sections for all indexer-* plugins were added, so they work out-of-the-box as you required in your comments. Also, it is not mandatory to specify fields for the actions (the schema is relaxed).

I included a new change, to avoid duplicate values in a field when someone tries to copy to the same field, like: 

```
<copy>
	<field source="title" dest="title"/>
</copy>
```

In addition, I added a new class (IndexWriterParams) to facilitate the process of obtaining and parsing values from the index-writers.xml file. Now, an instance of IndexWriterParams is passed to each IndexWriter instead of HashMap.

sebastian-nagel:
Hi @r0ann3l, thanks! I've continued testing, and was able to feed two Solr indexes in parallel. Great! Afaics, all requested changes have been made (also that of @lewismc).

To make the configuration work out of the box, I would suggest 3 changes:
- use only field names defined in the default schema.xml
`ERROR: [doc=http://nutch.apache.org/] unknown field 'search'
- default Solr core name should be "nutch" as described in the [tutorial](https://wiki.apache.org/nutch/NutchTutorial)

I've tried to fix these issues in "[a fork of NUTCH-1480](https://github.com/sebastian-nagel/nutch/commits/NUTCH-1480)". Feel free to cherry pick it from there.

I've also tried to make indexer-dummy work. Without success, the file is created but then overwritten:

- there are two instances of `IndexWriters` active, each having a separate instance of DummyIndexWriter.
   - the instance created from `IndexerOutputFormat.getRecordWriter(IndexerOutputFormat.java:39)` writes into the file
   - but later on the instance created from `IndexWriters.open(IndexWriters.java:187)` opens the file anew, at the end there is an empty file. Because it's two instances there is no possibility to check whether the file writer is already instantiated.

I see two potential solutions:
1. the IndexWriter interface method `open(job, name)` was defined with file indexers in mind (cf. NUTCH-1541/[CSVIndexWriter](https://github.com/sebastian-nagel/nutch/blob/NUTCH-1541/src/plugin/indexer-csv/src/java/org/apache/nutch/indexwriter/csv/CSVIndexWriter.java#L233)), an index writer can then decide to do nothing when called with name "commit".
2. do not call the `commit()` method explicitly (ev. also remove it from the interface: it does not safely work in distributed mode because it's not run in the reducers (see the comment in RabbitIndexWriter).

I tend to the second solution. It would also solve the problem of having two IndexWriters instances active. What do you think?

r0ann3l:
Hi @sebastian, thank you very much for your comments!!! I agree with your suggestions and I included the changes you propose from your fork.

About indexer-dummy, I also tried to make it work, but it was not possible. In theory, you can build as many instances of `IndexWriters` as you want, that you will always get the same instance, because it gotten from cache. So, the first issue I found was the `ObjectCache` uses the `Configuration` object itself as the key, and this object is not the same in each call. This causes that there are two instances of `IndexWriters` writing to same file, as you say. So, I replaced the key of `ObjectCache` with the UUID of the `Configuration` object.

Now, we have only one instance of `IndexWriters`, but there is another problem: when you try to commit the writers in `IndexingJob.index(IndexingJob.java151)` it is already closed from `IndexerOutputFormat.getRecordWriter(IndexerOutputFormat.java:44)`. Therefore, I moved the `commit()` call from `IndexingJob` to `IndexerOutputFormat`, just before the `close()` method is called.

I also, moved the indexers description from `IndexingJob` to `IndexerOutputFormat`, to avoid to build `IndexWriters` instance twice.

Thanks

sebastian-nagel:
Hi @r0ann3l,
did you verify that the change using `NutchConfiguration.getUUID(conf)` changes the behavior. Cf. [NUTCH-2407](https://issues.apache.org/jira/browse/NUTCH-2407?focusedCommentId=16180780&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16180780) which let me doubt it as it's a only random UUID for each Configuration object. 

Using the UUID in the ObjectCache makes the unit tests fail (TestGenerator): in fact the ObjectCache now returns the same object even if the configuration is different. We need actually really implement a hash value for Configuration objects.

lewismc:
@r0ann3l can you please update this PR inline with master?

r0ann3l:
Hi @sebastian-nagel:
In this case I propose to use an internal CACHE object, as the PluginRepository does, to store the IndexWriters object. The code could be something like this:
```
private static final WeakHashMap<String, IndexWriters> CACHE = new WeakHashMap<>();
public static synchronized IndexWriters get(Configuration conf) {
  String uuid = NutchConfiguration.getUUID(conf);
  if (uuid == null) {
    uuid = "nonNutchConf@" + conf.hashCode();
  }
  return CACHE.computeIfAbsent(uuid, k -> new IndexWriters(conf));
}
```
What do you think?

r0ann3l:
Hi @sebastian-nagel,

I changed the use of `ObjectCache.class` to an internal CACHE object, to avoid changing the behavior of other functionalities. In this case it is not necessary to have a different instance of `IndexWriters.class` for each `Configuration.class`. This is beacuse the index writer's configuration is handled in other individual file.

Also, I fixed an issue in `TestElasticIndexWriter.class` (associated with the use of `IndexWriterParams.class`), which causes the unit test fail.

sebastian-nagel:
Thanks, @r0ann3l! +1 - I've tested the solution in local and pseudo-distributed mode and was able to index into Solr (a single index). If there are no objections I'll commit/merge soon.

src/plugin/indexer-solr/src/java/org/apache/nutch/indexwriter/solr/SolrConstants.java
sebastian-nagel: It's still used for {{ZOOKEEPER_HOSTS = SOLR_PREFIX + "zookeeper.hosts"}}

PR #219, opened 2017-08-29
NUTCH-2415 : Create a JEXL based IndexingFilter

pipldev:
Add a new IndexingFilter that uses JEXL to decide whether to index a document.
The JEXL expression is read from the configuration property index.jexl.filter.


conf/nutch-default.xml
```diff
@@ -1566,6 +1566,16 @@ visit https://wiki.apache.org/nutch/SimilarityScoringFilter-->
   </description>
 </property>
 
+<!-- index-jexl-filter plugin properties -->
+
+<property>
+  <name>index.jexl.filter</name>
+  <value></value>
+  <description> A JEXL expression. If it evaluates to false,
```
jorgelbg: Probably we should add a comment here specifying which parameters are available for filtering. This would avoid checking the source code to see what you could use for filtering.

src/plugin/index-jexl-filter/src/java/org/apache/nutch/indexer/filter/JexlIndexingFilter.java
```diff
@@ -0,0 +1,119 @@
...
+public class JexlIndexingFilter implements IndexingFilter {
+
+	private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+	private Configuration conf;
+	private Expression expr;
+
+	@Override
+	public NutchDocument filter(NutchDocument doc, Parse parse, Text url, CrawlDatum datum, Inlinks inlinks) throws IndexingException {
+	    if (expr != null) {
```
jorgelbg: For what I see: `JexlUtil` returns null if it receives `null` or if cannot parse the expression:

https://github.com/apache/nutch/blob/b36692a6733e906be6fedbce8607397531666836/src/java/org/apache/nutch/util/JexlUtil.java#L72-L74

So if the expression cannot be parsed all the documents will be indexed because of this check, should we do a more strict check? Because since if you already enable the plugin you probably want your documents filtered (the plugin is not enabled by default).

src/plugin/index-jexl-filter/src/java/org/apache/nutch/indexer/filter/JexlIndexingFilter.java
```diff
@@ -0,0 +1,119 @@
...
+	        jcontext.set("contentMeta", metadataToContext(parse.getData().getContentMeta()));
+	        jcontext.set("parseMeta", metadataToContext(parse.getData().getParseMeta()));
+	        
+	        JexlContext context = new MapContext();
+	        for (Entry<String, NutchField> entry : doc) {
+	        	context.set(entry.getKey(), entry.getValue().getValues());
+	        }
+	        jcontext.set("doc", context);
+	                    
+			try {
```
jorgelbg: Do you mind applying the eclipse formatter to this block?

jorgelbg:
Probably would be a good idea to add a couple of tests to validate the implementation, testing an `IndexingFilter` is not very difficult you can check for instance the `index-basic` plugin. 

https://github.com/apache/nutch/blob/master/src/plugin/index-basic/src/test/org/apache/nutch/indexer/basic/TestBasicIndexingFilter.java

src/plugin/index-jexl-filter/src/java/org/apache/nutch/indexer/filter/JexlIndexingFilter.java
pipldev: On the other hand, if the indexer is enabled and configured, you probably want your documents indexed, right?
As a rule, I think plugins should not do anything when not configured (other than writing to log), but if there is a strong opposite opinion, I can change this.

jorgelbg: I partially agree with you, but in this case, the purpose of the plugin is to filter documents, so it has a really _active_ behaviour: if the wrong expression is entered you'll get an error logged but you will end with all the documents (that you wanted to exclude in the first place) in the index. 

I would prefer to get an error that the expression could not be parsed (probably due to a syntax error) of course this would cause the index step to fail. 

We could even go one step further: if the plugin is enabled we could treat the expression as a required parameter because otherwise, the plugin is not doing anything, this is done for instance on the `solr-indexer` plugin (will fail if the required parameters are not specified)

So we will not enable this plugin by default but if you enable the plugin you'll need to provide a valid expression.

pipldev: OK, I will throw a RuntimeException from setConf when null/invalid.

jorgelbg: 👍  the if expression on the `filter` method could also be removed then

pipldev:
All above comments have been fixed.

src/plugin/index-jexl-filter/src/test/org/apache/nutch/indexer/filter/TestJexlIndexingFilter.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.nutch.parse.ParseImpl;
+import org.apache.nutch.parse.ParseStatus;
+import org.apache.nutch.util.NutchConfiguration;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestJexlIndexingFilter {
+
+	@Test
+	public void testBasicIndexingFilter() throws Exception {
```
jorgelbg: Would be best to have a meaningful test name, something like `testAllowMatchingDocument` but in case that the test fails, will provide information about what the test is checking.

src/plugin/index-jexl-filter/src/test/org/apache/nutch/indexer/filter/TestJexlIndexingFilter.java
```diff
@@ -0,0 +1,80 @@
...
+		try {
+			NutchDocument result = filter.filter(doc, parse, new Text("http://nutch.apache.org/index.html"), crawlDatum, inlinks);
+			Assert.assertNotNull(result);
+			Assert.assertEquals(doc, result);
+		} catch (Exception e) {
+			e.printStackTrace();
+			Assert.fail(e.getMessage());
+		}
+		
+		doc.removeField("lang");
```
jorgelbg: This should be a separated test case something like `testBlockNotMatchingDocuments`. In tests is ok if we have a bit of repetition, but the idea is that each test contains one specific scenario (test case). 

It would be also great if you could add a test for when the expression is missing/invalid expression is entered, just checking that the proper exception get's triggered could do the trick.

src/plugin/index-jexl-filter/src/test/org/apache/nutch/indexer/filter/TestJexlIndexingFilter.java
```diff
@@ -0,0 +1,80 @@
...
+		doc.add("lang", "ru");
+		
+		try {
+			NutchDocument result = filter.filter(doc, parse, new Text("http://nutch.apache.org/index.html"), crawlDatum, inlinks);
+			Assert.assertNull(result);
+		} catch (Exception e) {
+			e.printStackTrace();
+			Assert.fail(e.getMessage());
```
jorgelbg: When this check is moved to a different test method you should use the exception assertion from JUnit, instead of the `try ... catch` block.

jorgelbg:
@pipldev Added a couple of comments about the test, but it's looking really good! Thanks for the contribution!

src/plugin/index-jexl-filter/src/test/org/apache/nutch/indexer/filter/TestJexlIndexingFilter.java
pipldev: I'm not sure I understand your comment. If I understand correctly, I should just remove the try...catch, and let the Exception be thrown, since that will fail the test with the exception.
For the new (third) test, where we want to create an exception, I assume we should use ExpectedException.
Agreed?
(BTW, all these problems with the test were also in the test you pointed me to as a template!)

jorgelbg: @pipldev Yes for testing the exception you could use something like

```java
@Test(expected = IndexOutOfBoundsException.class) 
public void empty() { 
     new ArrayList<Object>().get(0); 
}
```
The idea is that if you isolate this into a separated test you don't need the try ... catch because for that particular test you'll know if it's supposed to throw an exception or not (hence my suggestion of having separated tests). So for test `A` you know it's supposed to work as expected and you don't need to handle the exception because you know that when you're setting the data for the filter. And on test `B` you know that it's supposed to throw an exception because you made sure of that when configuring the filter and your assertions will reflect that. 

We try to follow the _boy scout rule_ when some code is touched/added we try to leave it in a better form than how we found it :). With this approach, we try to keep the stability of Nutch and make the life easier for the maintainers. And yes, some pieces of our code could use some love :)

lewismc:
@pipldev can you please bring this patch up-to-date and also consider deprecating some of the other IndexingFilters if you see fit. Thank you

YossiTamari:
Updated to resolve conflicts.
@lewismc How do I deprecate other IndexingFilters? Anyway, I don't think we should, since I see this more as an "power-user" filter, I think it's better if non-programmers can filter without learning JEXL and the relevant object model.

build.xml
```diff
@@ -1042,6 +1042,8 @@
         <source path="${plugins.dir}/index-geoip/src/java/" />
         <source path="${plugins.dir}/index-links/src/java/" />
         <source path="${plugins.dir}/index-links/src/test/" />
+        <source path="${plugins.dir}/index-jexl-filter/src/java/" />
```
lewismc: Can you also please make the entry for JavaDoc. You will find the entries elsewhere in build.xml

conf/nutch-default.xml
lewismc: Is this now addressed @jorgelbg ?

src/plugin/index-jexl-filter/src/java/org/apache/nutch/indexer/filter/JexlIndexingFilter.java
```diff
@@ -0,0 +1,126 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * An {@link org.apache.nutch.indexer.IndexingFilter} that allows filtering of
+ * documents based on a JEXL expression.
+ *
+ */
+public class JexlIndexingFilter implements IndexingFilter {
+
+	private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
lewismc: Can you please format all of the code with the [eclipse code formatter](https://github.com/apache/nutch/blob/master/eclipse-codeformat.xml)?

src/plugin/index-jexl-filter/src/test/org/apache/nutch/indexer/filter/TestJexlIndexingFilter.java
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class TestJexlIndexingFilter {
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+	@Test
```
lewismc: Again please format all code. Thank you

build.xml
sebastian-nagel: - the index-jexl-filter dir should also be added to the Eclipse target: there will be 4 entries in the main build.xml
- the plugin folders in build.xml are sorted lexicographically: index-jexl-filter should come before index-links
- please also add the package to plugins.index in default.properties. That's required to list the pluging on the [Javadoc overview page](https://builds.apache.org/job/nutch-trunk/javadoc/)
- also a package-info.java would be great

YossiTamari: @sebastian-nagel regarding default.properties: Is that the package name? Because the package name is org.apache.nutch.indexer.filter, which is already there. Should I change the package name?

sebastian-nagel: Yes, of course, the package name should be descriptive, e.g. org.apache.nutch.indexer.jexl. Thanks!

YossiTamari:
All above comments were addressed in the last commit.
Note that due to package renaming the history of this PR is a bit confusing.

sebastian-nagel:
+1 looks good

PR #224, opened 2017-09-26
Nutch 2433 - New configuration for HTML parser to keep the HTML nodes in outlinks metadata

maborec:
New configuration for HTML parser to keep the HTML nodes where outlinks are found, into their metadata.

### DIFF
--- a/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
+++ b/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
@@ -436,15 +436,16 @@ public class DOMContentUtils {
               try {
 
                 URL url = URLUtil.resolveURL(base, target);
-                outlinks.add(new Outlink(url.toString(), linkText.toString()
-                    .trim()));
-										
-                // NUTCH-2433 - Keep the node name where the URL was found into 
+                Outlink outlink = new Outlink(url.toString(), linkText
+                    .toString().trim());
+                outlinks.add(outlink);
+
+                // NUTCH-2433 - Keep the node name where the URL was found into
                 // the outlink metadata
                 if (keepNodenames) {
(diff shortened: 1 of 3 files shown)


lewismc:
I am +1 for committing to master. Tested locally and all good.

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
```diff
@@ -428,6 +438,15 @@ public void getOutlinks(URL base, ArrayList<Outlink> outlinks, Node node) {
                 URL url = URLUtil.resolveURL(base, target);
                 outlinks.add(new Outlink(url.toString(), linkText.toString()
                     .trim()));
+										
+                // NUTCH-2433 - Keep the node name where the URL was found into 
+                // the outlink metadata
+                if (keepNodenames) {
+                  MapWritable metadata = new MapWritable();
+                  metadata.put(new Text(srcTagMetaName), new Text(nodeName));
+                  outlinks.get(outlinks.size() - 1).setMetadata(metadata);
```
sebastian-nagel: Maybe keep a reference to the last outlink above to avoid the extra get operation. Only a matter of taste, of course.

sebastian-nagel:
The same change should be also applied to [src/plugin/parse-tika/.../DOMContentUtils.java](https://github.com/apache/nutch/tree/master/src/plugin/parse-tika/src/java/org/apache/nutch/parse/tika/DOMContentUtils.java#L403).

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
maborec: Good point, I agree. I've updated the code accordingly.

sebastian-nagel:
+1

PR #227, opened 2017-09-27
NUTCH-2436 Fix empty comment and unnecessary semicolon

kpm1985:


lewismc:
Thank you @kpm1985 if you are able to squash your commits in the future it would be greatly appreciated.
Thank you

PR #233, opened 2017-10-24
NUTCH-2449: Replace Tika LanguageIdentifier in language-identifier

YossiTamari:
Replaced with Optimaize LanguageDetector, which Tika has moved to as the default LanguageDetector.

### DIFF
--- a/src/plugin/language-identifier/build.xml
+++ b/src/plugin/language-identifier/build.xml
@@ -20,10 +20,6 @@
   <import file="../build-plugin.xml"/>
 
   <target name="init-plugin">
-    <echo>Copying language profiles</echo>
-    <copy todir="${build.classes}">
-      <fileset dir="${src.dir}" includes="**/*.ngp, **/*.properties"/>
-    </copy>
     <echo>Copying test files</echo>
     <copy todir="${build.test}">
       <fileset dir="${src.test}" includes="**/*.test, **/*.txt"/>
--- a/src/plugin/language-identifier/ivy.xml
+++ b/src/plugin/language-identifier/ivy.xml
@@ -36,6 +36,7 @@
   </publications>
 
   <dependencies>
+    <dependency org="com.optimaize.languagedetector" name="language-detector" rev="0.6"/>
   </dependencies>
   
 </ivy-module>
--- a/src/plugin/language-identifier/plugin.xml
+++ b/src/plugin/language-identifier/plugin.xml
@@ -25,6 +25,8 @@
       <library name="language-identifier.jar">
          <export name="*"/>
       </library>
+      <library name="language-detector-0.6.jar"/>
+      <library name="guava-18.0.jar"/>
    </runtime>
 
    <requires>
--- a/src/plugin/language-identifier/src/java/org/apache/nutch/analysis/lang/HTMLLanguageParser.java
+++ b/src/plugin/language-identifier/src/java/org/apache/nutch/analysis/lang/HTMLLanguageParser.java
@@ -16,10 +16,12 @@
  */
 package org.apache.nutch.analysis.lang;
 
+import java.io.IOException;
 // JDK imports
 import java.lang.invoke.MethodHandles;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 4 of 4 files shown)


YossiTamari:
If this pull request is accepted, the tests for the language-identifier plugin should be rewritten.

lewismc:
Hi @YossiTamari I think that this feature would be better proposed as a brand new plugin... the existing language-identifier plugin is stable and there is no real reason to get entirely rid of it. There is however always purpose in adding a new plugin... folks can then choose as they wish.
Can you repurpose this pull request to add a new plugin?

pipldev:
Sebastian made the same suggestion, but I really don't like it personally for two reasons:
1. 90% of the code of the plugin would be copied, and somebody would have to manage applying every fix to both plugins.
2. I think it would be really confusing to users to find two plugins that do exactly the same thing, with the only difference being an implementation detail.
Keep in mind that while the the existing plugin is stable, it relies on deprecated code, and presumably would break eventually when Tika decides to remove the deprecated implementation.

lewismc:
@pipldev thanks

> 90% of the code of the plugin would be copied, and somebody would have to manage applying every fix to both plugins.

I agree. There have been VERY few changes to this plugin in almost a decade. I don't anticipate that there will be too many changes moving forward.

> I think it would be really confusing to users to find two plugins that do exactly the same thing, with the only difference being an implementation detail.

Not if it is documented in a README.md contained within the plugin directory. Usually people interested in Linguistics are also very interested in the application or techniques used in tasks such as language identification, statistical machine translations, etc. i think any confusion could be easily avoided if we provide a simple README.

> Keep in mind that while the the existing plugin is stable, it relies on deprecated code, and presumably would break eventually when Tika decides to remove the deprecated implementation.

I agree, but as I said it is stable. We have tight integration between Nutch and Tika dev community. Once the code is removed over in Tika, it is trivial to remove the code over here.

At the end of the day it is down to the contributor if this code is to be updated or not, I just think preservation is an important part of what we do here. Retaining the legacy plugin is doing no harm whatsoever.
Thanks

YossiTamari:
Sorry for the mix-up, pipldev is my alternate account...

Speaking as a relatively new user of Nutch, I have never looked for a readme file in a plugin directory. It's a good thing, since I just sampled a few, and none had a readme file.

In my experience, getting into Nutch is a challenging task, and often requires looking into the code (and often debugging) to understand how things are supposed to work.
I do not think adding one more way to confuse users is a good idea. Just as an example for why it would be so confusing, the plugin has a number of configuration settings. Do we use the same properties to configure both plugins? Do we duplicate all of them in nutch-default.xml and try to explain in the comments there why there are two sets with the exact same functionalities?

sebastian-nagel:
Hi @YossiTamari, finally I've found the time to test the PR. Fetching your branch failed, to resolve conflicts I've created a [new branch](https://github.com/sebastian-nagel/nutch/tree/YossiTamari-NUTCH-2449) and applied your patch. One trivial fix: still need to copy `langmapping.properties` (used to parse HTML lang attribute) to runtime. Everything works fine! If there are no objection I'll merge soon. Thanks!

YossiTamari:
Thanks @sebastian-nagel, note that the tests still use org.apache.tika.language.LanguageIdentifier directly.

sebastian-nagel:
Ok, better to update the tests. Thanks for the hint, @YossiTamari!

lewismc:
@YossiTamari @sebastian-nagel this has been sitting for way too long. By the looks of the above correspondence the decision was made to overwrite the logic in language-identifier with the Optimaize logic.
[Tika LanguageDetector](https://tika.apache.org/2.1.0/api/index.html?org/apache/tika/language/detect/LanguageDetector.html) now offers several implementations which I think we could easily drive through configuration in nutch-default.xml.
Where do we want to go with this one? I can help with the effort.

sebastian-nagel:
Hi @lewismc, via NUTCH-2891/ b0cbea5 we already switched to the Optimaize language detector wrapped by Tika. I have it on my list to verify that this PR (and NUTCH-1397) is resolved en passant but hadn't the time to look over it. If you want, feel free to take over. Thanks!

lewismc:
OK @sebastian-nagel I will take it on yes. I'm nearly finished with my metrics documentation then I'll come back to this one.

lewismc:
Closing this PR off now in place of https://github.com/apache/nutch/pull/716

PR #235, opened 2017-10-24
Fix for NUTCH-2450 by Kenneth McFarland

kpm1985:


src/java/org/apache/nutch/parse/ParseOutputFormat.java
```diff
@@ -379,15 +378,16 @@ public static String filterNormalize(String fromUrl, String toUrl,
...
-          //FIXME: toDomain will never be null, correct?
           if (toDomain == null || toDomain.equals(origin)) {
             return null; // skip it
           }
         } else {
           String toHost = targetURL.getHost().toLowerCase();
-          //FIXME: toDomain will never be null, correct?
           if (toHost == null || toHost.equals(origin)) {
-            return null; // skip it
+            if (exemptionFilters == null // check if it is exempted?
```
sebastian-nagel: The [exemption filter|https://builds.apache.org/job/nutch-trunk/javadoc/org/apache/nutch/urlfilter/ignoreexempt/package-summary.html] is to define exemption for otherwise skipped external links. Should not be in the branch which handles internal links, otherwise we would have to redefine the exemption filter interface.

src/java/org/apache/nutch/parse/ParseOutputFormat.java
```diff
@@ -362,7 +362,6 @@ public static String filterNormalize(String fromUrl, String toUrl,
       if (ignoreExternalLinks) {
         if ("bydomain".equalsIgnoreCase(ignoreExternalLinksMode)) {
           String toDomain = URLUtil.getDomainName(targetURL).toLowerCase();
-          //FIXME: toDomain will never be null, correct?
```
sebastian-nagel: Please check whether URLUtil.getDomainName(URL url) may return null and add this information to the javadoc of the function. The FIXME should then be resolved accordingly.
If you're uncertain about the behavior, you could a test to TestURLUtil.testGetDomainDame() to test the actual behavior, e.g.,
```
    // test URL without host/authority
    url = new URL("file:/path/index.html");
    Assert.assertNull(URLUtil.getDomainName(url));
    Assert.assertNotNull(URLUtil.getDomainName(url));
    Assert.assertEquals("", URLUtil.getDomainName(url));
```
Of course, not all three assertions may succeed. Please also add such a test to your PR, or even more for edge cases of URLs without host or domain.

src/java/org/apache/nutch/parse/ParseOutputFormat.java
kpm1985: Thank you for this review @sebastian-nagel , I will be attacking this over this weekend. I appreciate the clarification in understanding. This project has absolutely fantastic reviewers, thank you again!

lewismc:
@kpm1985 ping, trying to close some issues here.

kpm1985:
@lewismc Apologies. Where would you like the unit test to be located?

lewismc:
[TestURLUtil](https://github.com/apache/nutch/blob/master/src/test/org/apache/nutch/util/TestURLUtil.java) testGetDomainDame() or additional tests for that test class.

kpm1985:
I closed the issue. The code is still labeled "FIXME", I don't want to mess up closing issues @lewismc  and since Eclipse has been being such a frustration issue I am just going to take a break from the time + frustration issues and come back. 

I hope that is ok, I have taken notes and will come back with a preemptive strike next time. I'm just not good with pressure right now. Thank you for giving me the direction, the next time I PR it will have more thought because of this.

PR #239, opened 2017-11-04
NUTCH-2442 Injector to stop if job fails to avoid loss of CrawlDb

Omkar20895:
- Added Job status checks in the classes: Injector, ReadHostDb, CrawlCompletionStats, ProtocolStatusStatistics, SitemapProcessor and DomainStatistics.

### DIFF
--- a/src/java/org/apache/nutch/util/ProtocolStatusStatistics.java
+++ b/src/java/org/apache/nutch/util/ProtocolStatusStatistics.java
@@ -122,8 +122,17 @@ public class ProtocolStatusStatistics extends Configured implements Tool {
     job.setNumReduceTasks(numOfReducers);
 
     try {
-      job.waitForCompletion(true);
-    } catch (Exception e) {
+      boolean success = job.waitForCompletion(true);
+      if(!success){
+        String message = jobName + " job did not succeed, job status: "
+            + job.getStatus().getState() + ", reason: "
+            + job.getStatus().getFailureInfo();
+        LOG.error(message);
+        // throw exception so that calling routine can exit with error
+        throw new RuntimeException(message);
+      }
+    } catch (IOException | InterruptedException | ClassNotFoundException e) {
+      LOG.error(jobName + " job failed", e);
       throw e;
     }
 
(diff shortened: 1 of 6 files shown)


src/java/org/apache/nutch/util/ProtocolStatusStatistics.java
```diff
@@ -122,8 +122,17 @@ public int run(String[] args) throws Exception {
     job.setNumReduceTasks(numOfReducers);
 
     try {
-      job.waitForCompletion(true);
-    } catch (Exception e) {
+      boolean success = job.waitForCompletion(true);
+      if(!success){
```
sebastian-nagel: Please use consistent and uniform formatting.

Omkar20895: @sebastian-nagel I did not understand, the formatting looks good to me. Can you please elaborate on what I am missing here? Thanks.

sebastian-nagel: There is a missing space before opening brackets/braces (also other occurrences). The formatting should follow the style defined by the Eclipse Code Formatter rules (https://github.com/apache/nutch/blob/master/eclipse-codeformat.xml). The easiest way is to import the rules into Eclipse (and other IDEs), in doubt, Eclipse allows to format the code from the command-line.

Omkar20895: Apologies for such a silly query 😅 . I use vim editor and everything lgtm, but, after importing the project into eclipse I could see the differences in the format. Thanks.

kpm1985:
This is somewhat of a tangent, but if we mavenize the build, we could use
some nice plugins for maven. Saw a few emails this week on this topic.

I looked into using nailgun and some other sweet things to ensure
consistency on a per commit basis, ie the maven import ordering plugin ( in
use under ASFLv2 in Fluo) for Log4J2.

Obviously, having each developer care enough to learn to import the style
into eclipse is good, a standard imo, but in the future we can automate
these styles. This is fantastic for formatting older class files etc..

Also, been lagging have lots of work but am serious about fixing old PR
(fyi for Sebastian)

Afaik ant is preferred less to maven.

sebastian-nagel:
Yes, that's clear. But the initial challenge is to port the complex build system to Maven.

PR #240, opened 2017-11-07
NUTCH-2456 - Redirected documents are not indexed

YossiTamari:
This is a defensive, minimal approach for fixing this issue.

### DIFF
--- a/src/java/org/apache/nutch/indexer/IndexerMapReduce.java
+++ b/src/java/org/apache/nutch/indexer/IndexerMapReduce.java
@@ -238,16 +238,16 @@ public class IndexerMapReduce extends Configured implements
     }
 
     // Whether to delete GONE or REDIRECTS
-    if (delete) {
-      if (fetchDatum != null && fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_GONE
+    if (delete && fetchDatum != null) {
+      if (fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_GONE
           || dbDatum != null && dbDatum.getStatus() == CrawlDatum.STATUS_DB_GONE) {
         reporter.incrCounter("IndexerStatus", "deleted (gone)", 1);
         output.collect(key, DELETE_ACTION);
         return;
       }
 
-      if (fetchDatum != null && fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_PERM
-          || fetchDatum != null && fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_TEMP
+      if (fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_PERM
+          || fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_TEMP
           || dbDatum != null && dbDatum.getStatus() == CrawlDatum.STATUS_DB_REDIR_PERM
           || dbDatum != null && dbDatum.getStatus() == CrawlDatum.STATUS_DB_REDIR_TEMP) {
         reporter.incrCounter("IndexerStatus", "deleted (redirects)", 1);
--- a/src/java/org/apache/nutch/indexer/IndexerMapReduce.java
+++ b/src/java/org/apache/nutch/indexer/IndexerMapReduce.java
@@ -238,18 +238,18 @@ public class IndexerMapReduce extends Configured implements
     }
 
     // Whether to delete GONE or REDIRECTS
-    if (delete && fetchDatum != null && dbDatum != null) {
-      if (fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_GONE
-          || dbDatum.getStatus() == CrawlDatum.STATUS_DB_GONE) {
+    if (delete) {
+      if (fetchDatum != null && fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_GONE
+          || dbDatum != null && dbDatum.getStatus() == CrawlDatum.STATUS_DB_GONE) {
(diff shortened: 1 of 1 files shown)


src/java/org/apache/nutch/indexer/IndexerMapReduce.java
```diff
@@ -256,20 +256,19 @@ public void reduce(Text key, Iterator<NutchWritable> values,
       }
```
sebastian-nagel: dbDatum is already used above (line 240 and following) when deleting gone pages and redirects. Deletions are done if either fetchDatum or dbDatum match the status (gone resp. redirect). Why not also relax these conditions so that dbDatum is optional? fetchDatum should exist, otherwise every index job will send deletions for **all** 404s/redirects in CrawlDb including those already deleted in the rounds before.

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
```diff
@@ -256,20 +256,19 @@ public void reduce(Text key, Iterator<NutchWritable> values,
...
 
-    if (fetchDatum == null || dbDatum == null || parseText == null
-        || parseData == null) {
+    if (fetchDatum == null || parseText == null || parseData == null) {
       return; // only have inlinks
     }
 
     // Whether to delete pages marked as duplicates
-    if (delete && dbDatum.getStatus() == CrawlDatum.STATUS_DB_DUPLICATE) {
+    if (delete && dbDatum!=null && dbDatum.getStatus() == CrawlDatum.STATUS_DB_DUPLICATE) {
```
sebastian-nagel: Code style: space around `!=`

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
```diff
@@ -309,23 +308,25 @@ public void reduce(Text key, Iterator<NutchWritable> values,
...
-        // If repr URL is excluded by filters we still accept this document
-        // but represented by its primary URL ("key") which has passed URL
-        // filters.
-        String urlString = filterUrl(normalizeUrl(url.toString()));
-        if (urlString != null) {
-          url.set(urlString);
-          fetchDatum.getMetaData().put(Nutch.WRITABLE_REPR_URL_KEY, url);
-        }
+      if (dbDatum!=null) {
+	      // Indexing filters may also be interested in the signature
```
sebastian-nagel: Code style: indentation 2 spaces per level, no tabs

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
YossiTamari: Fixed.

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
```diff
@@ -238,38 +238,37 @@ public void reduce(Text key, Iterator<NutchWritable> values,
     }
 
     // Whether to delete GONE or REDIRECTS
-    if (delete && fetchDatum != null && dbDatum != null) {
-      if (fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_GONE
-          || dbDatum.getStatus() == CrawlDatum.STATUS_DB_GONE) {
+    if (delete) {
```
sebastian-nagel: I would use
```
if (delete && fetchDatum != null) {
```
and skip further null checks of fetchDatum. In a continuous crawl the CrawlDb will hold many redirects and gone pages, significantly more than in the current segment which is indexed. Sending index deletions for all of them will slow down the index step.

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
```diff
@@ -238,38 +238,37 @@ public void reduce(Text key, Iterator<NutchWritable> values,
...
         return;
       }
 
-      if (fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_PERM
-          || fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_TEMP
-          || dbDatum.getStatus() == CrawlDatum.STATUS_DB_REDIR_PERM
-          || dbDatum.getStatus() == CrawlDatum.STATUS_DB_REDIR_TEMP) {
+      if (fetchDatum != null && fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_PERM
+          || fetchDatum != null && fetchDatum.getStatus() == CrawlDatum.STATUS_FETCH_REDIR_TEMP
+          || dbDatum != null && dbDatum.getStatus() == CrawlDatum.STATUS_DB_REDIR_PERM
```
sebastian-nagel: could simplify: `dbDatum != null && (status == redir_perm || status == redir_temp)`

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
YossiTamari: I will make the change, but isn't this code running only on the specific segments (and URLs) that were Fetched in this cycle?

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
YossiTamari: I think when looking at the whole condition, my code is more readable (as opposed to having nested parenthesis (like this))

src/java/org/apache/nutch/indexer/IndexerMapReduce.java
sebastian-nagel: Yes, but the indexer takes also the full CrawlDb (and optionally LinkDb) as input. In a crawl with many cycles the CrawlDb may become significantly bigger than the currently processed segment(s). There is no way to filter CrawlDb items without counter-parts in the segments beforehand, all end up in the reducer.

PR #243, opened 2017-11-20
NUTCH-2463 - Enable sampling CrawlDB

YossiTamari:
Added the parameter -sample <float>. The float should be a fraction, only this fraction of records will be processed.


sebastian-nagel:
+1

Good idea. Patch/PR looks good to me.

YossiTamari:
Thanks!

PR #244, opened 2017-11-23
Fix for NUTCH-2464 get textual content from nested heading nodes

jorgelbg:
As suggested by @sebastian-nagel refactored the `getNodeValue` method to use `NodeWalker` iterators. This allows traversing the entire DOM tree in case of nested nodes as explained on the issue. Added a test case for this issue as well.


sebastian-nagel:
Looks good to me. Thanks, @jorgelbg!

PR #247, opened 2017-11-27
Nutch 2465

okedoki:
Please verify the request carefully.  I think about possible pitfalls.


sebastian-nagel:
Hi @okedoki, thanks! Could you please first
- remove all files from the pull request which shouldn't be there: `*~`, .classpath, .project, etc. If you add a new file there must be a clear reason for it.
- choose meaningful commit messages which state what has been changed. It's good to reference the NUTCH-XXX issue but an additional description (e.g., "moved classes into package o.a.n.protocol.interactiveselenium.handlers following directory structure") will help to review and also makes it easier to browse the history later on
- the author of a commit is tracked by git, no need to repeat it in the message

The classpath issues need to be fixed in build.xml ("target: eclipse").

okedoki:
Thanks for your comment.
Unfortunately, I found out some bugs with this solution. I'm closing this pull request.
Feel free to propose a (working) solution.

PR #268, opened 2017-12-25
Improved base config for nutch

dhirajforyou:


dhirajforyou:
My mistake, raised PR against the apache/nutch instead of my forked trunk.
Please decline the PR. Its related for testing purpose only.

PR #271, opened 2018-01-03
NUTCH-2492: Add more configuration parameters to crawl script

mfeltscher:
Made the following configuration options configurable using command arguments:
- numSlaves
- numTasks
- sizeFetchlist
- timeLimitFetch
- numThreads

I also refactored the usage output in order to make it better readable.

PR #273, opened 2018-01-09
NUTCH-2493: Add new parameter to crawler script to configure sitemap processing frequency

mfeltscher:
While using the crawler script with the sitemap processing feature introduced in #271 I encountered some performance issues when working with large sitemaps.
Therefore one should be able to specify if sitemap processing based on HostDB should take place and if so how frequently it should be done.

### DIFF
--- a/src/bin/crawl
+++ b/src/bin/crawl
@@ -63,6 +63,37 @@ function __to_seconds() {
   echo $SECONDS
 }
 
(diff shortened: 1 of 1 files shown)


lewismc:
I am +1 here, tested on local test crawl.

sebastian-nagel:
+1, patch looks good

PR #275, opened 2018-01-12
NUTCH-2495: Use -deleteGone instead of clean job in crawler script while indexing

mfeltscher:
Instead of running `bin/nutch` clean after indexing the documents run `bin/nutch index` with the `-deleteGone` flag which instead of just deleting gone and duplicated documents also deletes redirects from the index.


sebastian-nagel:
The changes in IndexerMapReduce are already contained in [NUTCH-2456](https://issues.apache.org/jira/browse/NUTCH-2456). I've replaced this PR by PR #517 which includes the changes in src/bin/crawl. Thanks for your contribution, @mfeltscher!

PR #278, opened 2018-01-17
NUTCH-2481 HostDatum deltas(previous step statistics) and Metadata expressions

okedoki:
The logic of updatehostdb is changed slightly.

In case of specification of hostdb.deltaExpression, we dont reset statistics in mapper, but send the previous step statistic first to the reducer and reset it afterwards.

In line 215 of the mapper  
 if (readingCrawlDb)
is replaced by  
  if (readingCrawlDb && !isDeltaStatisticCalculated) {
         hostDatum.resetStatistics();
-      }

Please, verify that logic doesn't break the current functionality.


src/java/org/apache/nutch/hostdb/UpdateHostDbReducer.java
```diff
@@ -88,95 +97,128 @@ public void configure(JobConf job) {
...
       }
     }
 
+    stringDeltaExpression = job
+        .get(UpdateHostDb.HOSTDB_UPDATEDB_DELTA_EXPRESSION);
+    if (!org.apache.commons.lang3.StringUtils.isEmpty(stringDeltaExpression)) {
+      // Create or retrieve a JexlEngine
+      JexlEngine jexl = new JexlEngine();
+
+      // Dont't be silent and be strict
```
YossiTamari: `Dont't` must be a typo, but beyond that, `setSilent(true)` seems to contradict this comment.

okedoki: @YossiTamari 
The funny part it is a copy-paste from ReadHostDb line 83.
How do you propose to fix it?

YossiTamari: I'm not sure what the original intent was, but maybe we should replace this whole code with: 
`this.deltaExpression = org.apache.nutch.util.JexlUtil.parseExpression(stringDeltaExpression);`?

PR #281, opened 2018-01-23
NUTCH-2503: Add option to run tests for a single plugin

mfeltscher:
Allows running tests for a given plugin as follows: `ant test-plugin -Dplugin=<plugin>`
Example: `ant test-plugin -Dplugin=any23`


nmaro:
LGTM

PR #282, opened 2018-01-27
fix for NUTCH-2505 contributed by lianyuzhe

lianyuzhe:
fix the problem of nutch does not delete the .locked file when the generator partition got an exception.

### DIFF
--- a/src/java/org/apache/nutch/crawl/Generator.java
+++ b/src/java/org/apache/nutch/crawl/Generator.java
@@ -752,6 +752,7 @@ public class Generator extends NutchTool implements Tool {
       }
     } catch (Exception e) {
       LOG.warn("Generator: exception while partitioning segments, exiting ...");
+      LockUtil.removeLockFile(getConf(),lock);
       fs.delete(tempDir, true);
       return null;
     }


src/java/org/apache/nutch/crawl/Generator.java
```diff
@@ -752,6 +752,7 @@ public Generator(Configuration conf) {
       }
     } catch (Exception e) {
       LOG.warn("Generator: exception while partitioning segments, exiting ...");
+      LockUtil.removeLockFile(getConf(),lock);
```
sebastian-nagel: Thanks @lianyuzhe, good catch!
+1 to remove the lock of the CrawlDb. There may be good or broken segments left from the partition jobs run in the loop. Would be nice to clean them up as well. But that's clearly a different task.

lianyuzhe: do you think it is a good way to handle the Exception in the partitionSegment function? We can delete the segments is there is an Exception, and let the partition jobs continue.

sebastian-nagel: Yes, of course, if it's possible to properly catch the exception and safely delete the generated segment(s). It's your decision whether to extend the PR or to open a new issue. Thanks!

PR #285, opened 2018-02-12
NUTCH-1749 Optionally exclude title from content field

steeveb972:
Hello,

Following the description and the comment provided in this jira, I propose a patch to add the ability to exclude some tags in HTML and Tika parsers.

Regards,

Steeve

### DIFF
--- a/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
+++ b/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
@@ -19,10 +19,7 @@ package org.apache.nutch.parse.html;
 
 import java.net.URL;
 import java.net.MalformedURLException;
-import java.util.Collection;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Stack;
+import java.util.*;
 
 import org.apache.nutch.parse.Outlink;
 import org.apache.nutch.util.NodeWalker;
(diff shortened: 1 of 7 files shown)


src/plugin/parse-html/src/java/org/apache/nutch/parse/html/DOMContentUtils.java
```diff
@@ -116,36 +114,50 @@ public void setConf(Configuration conf) {
...
   }
 
   // returns true if abortOnNestedAnchors is true and we find nested
   // anchors
   private boolean getTextHelper(StringBuffer sb, Node node,
-      boolean abortOnNestedAnchors, int anchorDepth) {
+                                boolean abortOnNestedAnchors, int anchorDepth, Set<String> excludedElementNames) {
     boolean abort = false;
     NodeWalker walker = new NodeWalker(node);
+    Set<String> lcExcludedElementNames = new HashSet<>();
```
jorgelbg: We should avoid duplicating the exclusion set. This method is executed many times. We could use a `TreeSet` when it is invoked, delegating the comparison to the `Set` implementation.

src/plugin/parse-tika/src/java/org/apache/nutch/parse/tika/DOMContentUtils.java
```diff
@@ -107,50 +103,65 @@ public void setConf(Configuration conf) {
   }
 
   /**
-   * This method takes a {@link StringBuffer} and a DOM {@link Node}, and will
+   * This method takes a {@link StringBuffer}, a DOM {@link Node}
+   * and an excluded element {@link Set}, and will
    * append all the content text found beneath the DOM node to the
-   * <code>StringBuffer</code>.
-   * 
+   * <code>StringBuffer</code> without the mentioned element names in the <code>Set</code>.
```
jorgelbg: Same as the previous comment.

jorgelbg (changes requested):
@steeveb972 Thanks for your contribution! 
I've added some comments about your implementation that we should address before merging. Also, we need to fix the conflicts with the master branch.

lewismc:
Hi @steeveb972 this seems like a really useful patch. Are you able to update it and we try to get it into master?

steeveb972:
Hi, it has been a while ;-) I will have a look at it this week-end

PR #286, opened 2018-02-14
Nutch 2510

okedoki:
Now host db is not created automatically in the crawl script, you can use hostdb for generate plut the description is added.


okedoki:
Now we check existence of hostdb directroy before we process
  echo "Generating a new segment"  
  if [[ "$HOSTDBGENERATE" = true  && -d "$CRAWL_PATH"/hostdb ]]; 

One question left,  how should we modify(or shouldn't) this part?
 
if [[ "$SITEMAPS_FROM_HOSTDB_FREQUENCY" = "always" || ("$SITEMAPS_FROM_HOSTDB_FREQUENCY" = "once" && $a -eq 1) ]]; then
    # create / update HostDB on first run
    [[ $a -eq 1 ]] && __update_hostdb

    # sitemap processing based on HostDB
    if [[ -d "$CRAWL_PATH"/hostdb ]]; then
      echo "Processing sitemaps based on hosts in HostDB"
      __bin_nutch sitemap "$CRAWL_PATH"/crawldb -hostdb "$CRAWL_PATH"/hostdb -threads $NUM_THREADS
    fi
  fi

PR #290, opened 2018-03-06
NUTCH-2522  Bidirectional URL exemption filter

okedoki:


conf/db-ignore-external-exemptions-bidirectional.txt
```diff
@@ -0,0 +1,33 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
sebastian-nagel: Configuration files should be added as *.template. And are "instantiated" (copied) during the first compilation. Users than can modify the content without conflicts and undesired overwrites.

conf/db-ignore-external-exemptions-bidirectional.txt
```diff
@@ -0,0 +1,33 @@
...
+# Each non-comment, non-blank line contains a regular expression
+# prefixed by '+' or '-'.  The first matching pattern in the file
+# determines whether a URL is exempted or ignored.  If no pattern
+# matches, the URL is ignored.
+
+
+
+# Example 1:
+#----------
+# To exempt urls ending with image extensions, uncomment the below line
```
sebastian-nagel: Description does not fit the following line/rule.

conf/db-ignore-external-exemptions-bidirectional.txt
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#
+#
+# Exemption rules to db.ignore.external.links
+
+
+# Format :
+#--------
+# The format is same same as `regex-urlfilter.txt`.
+# Each non-comment, non-blank line contains a regular expression
```
sebastian-nagel: The description does not match the implementation.

conf/db-ignore-external-exemptions-bidirectional.txt
```diff
@@ -0,0 +1,33 @@
...
+# prefixed by '+' or '-'.  The first matching pattern in the file
+# determines whether a URL is exempted or ignored.  If no pattern
+# matches, the URL is ignored.
+
+
+
+# Example 1:
+#----------
+# To exempt urls ending with image extensions, uncomment the below line
+-(www.)
```
sebastian-nagel: Why the rule starts with `+-`? The dot is not escaped, would also apply to `wwwfinder` catching `wwwf`.

sebastian-nagel (changes requested):
Clean PR with correct code format and documentation! Thanks, @okedoki!

Afaics, the implementation does the following:
1. take the lowercased host part of both from and to URL
2. match all regex rules defined in the rules files and remove the matched part
3. finally, if from and to host are equal return true => URL is accepted ("exempted" from ignore external host exclusion)

Is this correct?

Wouldn't be a different rule file format more suitable?
- the leading +/- is not used
- don't know whether this makes sense, but could also define the replace string, ev. including references to captured groups, cf. the file format used by urlnormalizer-regex

okedoki:
@sebastian-nagel 
Thanks for the suggestion to use urlnormalizer-regex. I rewrote the plugin based on this approach( now it makes sense to refactor urlnormalizer-regex and this plugin to use the same code base).

The usage is correct, at this moment we apply the same regex for both input and output url and see if they match each other. 

In the future it can be improved with two separated regex for input and output.

PR #293, opened 2018-03-08
NUTCH-2517 mergesegs corrupts segment data

lewismc:
This is mostly a cleanup of the Classes concerned with https://issues.apache.org/jira/projects/NUTCH/issues/NUTCH-2517, namely SegmentMerger and LinkDb


lewismc:
Ni semantic changes here folks, will commit in 24hrs unless objections.

PR #295, opened 2018-03-14
NUTCH-2516 Hadoop imports use wildcards

lewismc:
Large but simple patch which addresses https://issues.apache.org/jira/projects/NUTCH/issues/NUTCH-2516


lewismc:
I would like to commit by EoB today if possible. We can then shift priority to #299 I will that through with @kpm1985 

Any comments folks?

.gitignore
```diff
@@ -6,3 +6,8 @@ build/
 runtime/
 logs/
 /bin/
+.classpath
+.naivebayes-model.crc
+.project
+ivy/ivy-2.3.0.jar
```
sebastian-nagel: What about ivy/ivy-2.4.0.jar which has been committed in f9981d4? It's supposed to be "installed" by `ant ivy-download`.

lewismc: Thanks @sebastian-nagel I've updated the PR to also prevent ```ivy-2.4.0.jar``` from being added to SCM.

sebastian-nagel: +1 Maybe also remove this file to avoid confusions?

sebastian-nagel: and please go forward and commit, thanks!

PR #305, opened 2018-03-26
NUTCH-2447 Work-around SSLProtocolException: handshake alert: unrecognized_name

sebastian-nagel:
- apply Markus' patch
- if reconnect fails throw HTTPException

This solution (work-around) has been tested in production.


lewismc:
+1

PR #306, opened 2018-03-27
NUTCH-2545 Upgrade to Any23 2.2

lewismc:
This issue addresses https://issues.apache.org/jira/projects/NUTCH/issues/NUTCH-2545 and passes all unit tests.
I have a feeling we can take the upgrade further and remove the [SAX fix logic](https://github.com/apache/nutch/blob/master/src/plugin/any23/src/java/org/apache/nutch/any23/Any23ParseFilter.java#L111-L116) previously implemented as this should be fixed in Any23 2.2. 

@HansBrende wdyt?


HansBrende:
@lewismc I've figured out all the details of the bugs.

Documented here in [ANY23-340](https://issues.apache.org/jira/projects/ANY23/issues/ANY23-340).

lewismc:
boom, for the time being we can upgrade to Any23. When the upgrade to 2.3 comes, then we can remove conditional logic.

HansBrende:
@lewismc that's fine, but beware: the SAX pre-processor actually screws up a lot of triples! 

In the microdata test, that 39th triple extracted with the SAX pre-processor is actually a *bug* (see ANY23-340). 

Also, the SAX pre-processor removes all the namespaces specified in the html element. Which means that, even though the BBC test file specified the namespaces xmlns:og="http://opengraphprotocol.org/schema/" and  xmlns:rnews="http://iptc.org/std/rNews/2011-10-07#", the extractors ignored those, defaulting to "http://ogp.me/ns#" for the "og" namespace, and "rnews:" for the "rnews" namespace.

lewismc:
@HansBrende I removed the previous SAX logic due to the concerns viewed above. We can revisit when we release Any23 2.3

PR #307, opened 2018-03-27
NUTCH-2518 Cleaning up the file system after a job failure.

Omkar20895:
- Added a new method cleanupAfterFailure which was already existing in Injector.
- I have added this only in one of the 25 instances of job.waitForCompletion(true), please go ahead and review it and I will incorporate the review comments in the upcoming commits/changes.

Confusion: 
- When a job failure happens do we need to do System.exit(non-zero status) or throw a runtime exception with a custom message specific to each class? I went ahead with the later one. Please suggest otherwise with reason. 

Thanks.


src/java/org/apache/nutch/crawl/CrawlDb.java
```diff
@@ -129,17 +129,23 @@ public void update(Path crawlDb, Path[] segments, boolean normalize,
...
+        cleanupAfterFailure(outPath, lock, fs);
+        throw new RuntimeException(message);
+      }
     } catch (IOException | InterruptedException | ClassNotFoundException e) {
-      FileSystem fs = crawlDb.getFileSystem(getConf());
-      LockUtil.removeLockFile(fs, lock);
-      Path outPath = FileOutputFormat.getOutputPath(job);
-      if (fs.exists(outPath))
-        fs.delete(outPath, true);
+      LOG.error("Crawl job failed ", e);
```
lewismc: Please always use parameterized logging e.g.
```LOG.error("Crawl job failed {}", e);```

Comment (Omkar20895) [reply src/java/org/apache/nutch/crawl/CrawlDb.java]: ACK! I will incorporate this change and send a commit covering all the 25 instances. Thanks.

Comment (sebastian-nagel) [reply src/java/org/apache/nutch/crawl/CrawlDb.java]: +1

The CrawlDb cleanup method looks same as that from Injector: maybe move it to CrawlDb - there are already methods to lock and install the CrawlDb shared by various jobs (inject, update, merge, generate).

Comment (Omkar20895) [reply src/java/org/apache/nutch/crawl/CrawlDb.java]: Why not shift the sub routine cleanup to the class nutch/src/java/org/apache/nutch/util/NutchJob.java? This is a nutch job utils file, I think this would be more appropriate. I have cross checked that cleanup sub routine would be useful in outside crawld like in hostdb, indexer etc.

**Comment by Omkar20895** [comment]
I have added a new commit, please feel free to review it and let me know if any changes are required. I will squash the commits at the end.

File: src/java/org/apache/nutch/crawl/Injector.java [hunk 2]
```diff
@@ -420,7 +420,7 @@ public void inject(Path crawlDb, Path urlDir, boolean overwrite,
             + job.getStatus().getState() + ", reason: "
             + job.getStatus().getFailureInfo();
         LOG.error(message);
-        cleanupAfterFailure(tempCrawlDb, lock, fs);
+        CrawlDb.cleanupAfterFailure(tempCrawlDb, lock, fs);
```
sebastian-nagel: should be `NutchJob.cleanupAfterFailure(...)`

src/java/org/apache/nutch/crawl/Injector.java
```diff
@@ -463,19 +463,7 @@ public void inject(Path crawlDb, Path urlDir, boolean overwrite,
...
-
-  public void cleanupAfterFailure(Path tempCrawlDb, Path lock, FileSystem fs)
-      throws IOException {
-    try {
-      if (fs.exists(tempCrawlDb)) {
-        fs.delete(tempCrawlDb, true);
-      }
-      LockUtil.removeLockFile(fs, lock);
-    } catch (IOException e) {
+      CrawlDb.cleanupAfterFailure(tempCrawlDb, lock, fs);
```
sebastian-nagel: same as above

src/java/org/apache/nutch/crawl/Generator.java
```diff
@@ -838,12 +846,21 @@ public Generator(Configuration conf) {
...
       try {
-        int complete = job.waitForCompletion(true)?0:1;
+        boolean success = job.waitForCompletion(true);
+        if (!success) {
+          String message = "Generator job did not succeed, job status:"
+              + job.getStatus().getState() + ", reason: "
+              + job.getStatus().getFailureInfo();
+          LOG.error(message);
+          NutchJob.cleanupAfterFailure(tempDir, lock, fs);
+          NutchJob.cleanupAfterFailure(tempDir2, lock, fs);
```
sebastian-nagel: ok, it's not an error to try to remove the lock file a second time.

src/java/org/apache/nutch/crawl/Injector.java
Omkar20895: I will change this, Thanks.

src/java/org/apache/nutch/crawl/Generator.java
Omkar20895: No @sebastian-nagel. AFAIK if I try to remove a lock file that does not exist, the function would just return False.

sebastian-nagel:
+1 Thanks, @Omkar20895! I'll step forward and commit the fix.

Omkar20895:
Thanks @sebastian-nagel :)

sebastian-nagel:
Squashed, rebased and committed in 8682b96. Thanks again, @Omkar20895!

lewismc:
This is ready to be merged @sebastian-nagel ?

PR #312, opened 2018-04-11
NUTCH-2533 Injector: NullPointerException if seed URL dir contains non-file entries

sebastian-nagel:
- read directory explicitly and log all non-file entries
- exit early if no seed URL files are present


lewismc:
+1

PR #325, opened 2018-04-23
NUTCH-2571 SegmentReader -list fails to read segment

sebastian-nagel:
- fix type of value (CrawlDatum not Text)


Omkar20895:
+1

PR #328, opened 2018-05-09
NUTCH-2576 HTTP protocol implementation based on okhttp

sebastian-nagel:
A Nutch protocol plugin based on [okhttp](http://square.github.io/okhttp/):

- derived from @jnioche's implementation for [storm-crawler#443](/DigitalPebble/storm-crawler/issues/443)
  - use okhttp's internal buffer for buffering content
- adapted to be compatible to Nutch and behave almost the same as protocol-http
  - moved shared configuration settings to HttpBase (lib-http)
- unit tests taken from protocol-http

TODOs:
- verify that issues reported in NUTCH-2549 do not appear again
- complete unit tests
- benchmark and large-scale test

For HTTP/2 support a Java version at runtime is needed which supports ALPN:

```
export NUTCH_JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64

bin/nutch parsechecker \
   -Dplugin.includes='protocol-okhttp|parse-html' \
   -Dhttp.useHttp2=true \
   -Dstore.http.headers=true \
   https://www.google.com/
...
_response.headers_=HTTP/2 200
date: Wed, 09 May 2018 11:23:21 GMT
expires: -1
cache-control: private, max-age=0
content-type: text/html; charset=ISO-8859-1
p3p: CP="This is not a P3P policy! See g.co/p3phelp for more info."
content-encoding: gzip
server: gws
x-xss-protection: 1; mode=block
...
```

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -215,6 +215,7 @@
       <packageset dir="${plugins.dir}/protocol-http/src/java"/>
       <packageset dir="${plugins.dir}/protocol-httpclient/src/java"/>
       <packageset dir="${plugins.dir}/protocol-interactiveselenium/src/java"/>
+      <packageset dir="${plugins.dir}/protocol-okhttp/src/java"/>
       <packageset dir="${plugins.dir}/protocol-selenium/src/java"/>
       <packageset dir="${plugins.dir}/publish-rabbitmq/src/java"/>
       <packageset dir="${plugins.dir}/scoring-depth/src/java"/>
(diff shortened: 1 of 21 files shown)


jnioche:
@sebastian-nagel one thing I noticed with OkHttp is that its [ConnectionPool](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/ConnectionPool.java) (default maxIdle 5, with eviction after 5 mins)  struggles when used with many threads and different hostnames, which would typically be the case with Nutch (and StormCrawler). I have seen an average of 1.5s and up to 6s contention on the ConnectionPool, my guess is that [the cleanup method](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/ConnectionPool.java#L199) and its synchronized block is the main culprit. It iterates on all the connections but removes only the one which has been idle for the longest.

Apart from that okHTTP is great: pretty robust and less arcane than Apache HTTPClient IMHO.

sebastian-nagel:
Done:
- large-scale test (distributed mode on CDH 5.14.2): 195 million pages fetched from 90 million hosts using 48 Fetcher tasks each with 120 threads. No issues with the connection pool, at least, not remarkable unless the waits for locks described or linked in [NUTCH-2578](https://issues.apache.org/jira/browse/NUTCH-2578) are addressed.

New TODOs:
- setting/using Cookies
- re-throw exceptions as HttpException so that they can be handled by Fetcher

PR #329, opened 2018-05-10
NUTCH-2562 protocol-http fails to read large chunked HTTP responses

sebastian-nagel:
- if http.content.limit is reached skip remaining chunked content
  including any headers in the trailer


GerardBouchar:
Is there something missing before this can be merged ?

sebastian-nagel:
No. It's only a good practice to wait for a review and at least one +1 because of the principle "two heads are better than one".

PR #340, opened 2018-06-04
Fixes for NUTCH-2412 contributed by r0ann3l

r0ann3l:
The exchange component acts in indexing job and decides which index writer a document should go to.
Includes:
1. Extension point.
2. Plugin based on JEXL expressions.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -177,6 +177,7 @@
       <packageset dir="${plugins.dir}/creativecommons/src/java"/>
       <packageset dir="${plugins.dir}/feed/src/java"/>
       <packageset dir="${plugins.dir}/headings/src/java"/>
+      <packageset dir="${plugins.dir}/exchange-jexl/src/java"/>
       <packageset dir="${plugins.dir}/index-anchor/src/java"/>
       <packageset dir="${plugins.dir}/index-basic/src/java"/>
       <packageset dir="${plugins.dir}/index-geoip/src/java"/>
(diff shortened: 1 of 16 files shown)


sebastian-nagel:
Hi @r0ann3l,
could you add package-info.java files to the packages org.apache.nutch.exchange (main) and org.apache.nutch.exchange.jexl (plugin)? Just one or two sentences, so that on the [API overview page](https://builds.apache.org/job/Nutch-trunk/javadoc/) the packages and the plugin are explained. Thanks!

If no exchange plugin is configured in plugin.includes, the indexer exits with a NPE:
```
Indexer: java.lang.NullPointerException
        at org.apache.nutch.exchange.Exchanges.<init>(Exchanges.java:68)
        at org.apache.nutch.indexer.IndexWriters.<init>(IndexWriters.java:103)
        at org.apache.nutch.indexer.IndexWriters.lambda$get$0(IndexWriters.java:60)
        at java.util.Map.computeIfAbsent(Map.java:957)
        at org.apache.nutch.indexer.IndexWriters.get(IndexWriters.java:60)
        at org.apache.nutch.indexer.IndexingJob.index(IndexingJob.java:123)
        at org.apache.nutch.indexer.IndexingJob.run(IndexingJob.java:235)
        at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:70)
        at org.apache.nutch.indexer.IndexingJob.main(IndexingJob.java:244)
```
I would opt to keep the use of exchange plugins optional:
- it's a plugin, hence optional. If none is configured, just route all documents to all index writers. It's ok, to log a warning about this.
- always keep the simplest use case (a single index writer) as simple as possible. It should be possible to do this without configuring any exchange plugin.

r0ann3l:
Hi @sebastian-nagel 
Thanks for your review. I made a new commit where the exception you mentioned is fixed. I added the package-info.java files too.

Talking about the optional use of exchange plugins, I agree with you and I modified the behavior to be optional. But I kept the default exchange (this one indicates where the documents should be routed when the they don't match any of the others exchanges) and only in the case there is no available exchanges (even the default) the documents will be sent to all index writers.

sebastian-nagel:
+1 lgtm. Also tested to index into a single Solr index using the default configuration. Feel free to commit!

PR #349, opened 2018-06-13
NUTCH-2597: fixed cleanup()

sju:


src/java/org/apache/nutch/hostdb/UpdateHostDbReducer.java
```diff
@@ -74,6 +74,7 @@
     *
     * @param job
     */
+  @Override
```
sebastian-nagel: +1
This issue is the best example why @Override should be used: to avoid that by accident a method has not the intended signature.

src/java/org/apache/nutch/indexer/CleaningJob.java
```diff
@@ -64,10 +64,12 @@ public void setConf(Configuration conf) {
       Mapper<Text, CrawlDatum, ByteWritable, Text> {
     private ByteWritable OUT = new ByteWritable(CrawlDatum.STATUS_DB_GONE);
 
+    @Override
     public void setup(Mapper<Text, CrawlDatum, ByteWritable, Text>.Context context) {
     }
 
-    public void cleanup() throws IOException {
+    @Override
+    public void cleanup(Context context) throws IOException {
```
sebastian-nagel: Could also remove the method implementation. The superclass Reducer already implements already a do-nothing cleanup(context).

If you have time: there are a couple of other cleanup() methods without the context argument. Probably same mistake but harmless as they "do nothing". Need to check in detail but `git grep -A2 'cleanup()'` finds a couple of them. Thanks, @sju!

sju:
Did some cleanup ;-)

sebastian-nagel:
Tested fix: the exception in the ResolverThreads is gone. Thanks, @sju!

PR #351, opened 2018-06-21
fix for NUTCH-2600: indexer-solr refactoring

r0ann3l:
This patch includes an update for solrj library from version 5.5.0 to 7.3.1. According to [SOLR-8903](https://issues.apache.org/jira/browse/SOLR-8903), the lastest available version of this library doesn't include `DateUtil` class, so for formatting the dates I used `DateTimeFormatter.ISO_INSTANT.format(d.toInstant())` instead. In addition, the weight of a document is no longer supported by this library as an attribute of `SolrInputDocument` class and for that reason I include a new paramater: "weight.field", indicating the field's name where the document weight should be written. Also, Basic Authentication is now supported.

On the other side, I refactored the code, deleting some unused constants, unsed parameters and including others in index-writers.xml.

This PR was tested in Solr 6.3.0 and 7.3.1 in standalone and cloud mode. Basic Authentication was tested in these versions too.

### DIFF
--- a/conf/index-writers.xml.template
+++ b/conf/index-writers.xml.template
@@ -22,10 +22,10 @@
   <writer id="indexer_solr_1" class="org.apache.nutch.indexwriter.solr.SolrIndexWriter">
     <parameters>
       <param name="type" value="http"/>
-      <!-- Solr URL (default core name is "nutch" but you may change it): -->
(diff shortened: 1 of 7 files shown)


sebastian-nagel:
Lgtm! Successfully tested with Solr 6.6.0. Thanks, @r0ann3l!

PR #354, opened 2018-06-25
NUTCH-2611: Add line-breaks when parsing HTML block-level elements

YossiTamari:
When the configuration property parser.html.line.separators contains a list of tags, a newline is added before and after the text content of this tag.


sebastian-nagel:
+1 lgtm. The plain-text layout is now indeed more readable - line breaks after head lines, <p>, etc. Will commit soon if there are no objections. Thanks, @YossiTamari!

PR #356, opened 2018-06-29
fix for NUTCH-2602: Index writers description

r0ann3l:
When the index writers description is shown, it appears in the form **\<key>:\<value>** where **\<key>** is the parameter's name used on index-writers.xml file and **\<value>** is the value that the parameter has. To acomplish this, the `describe()` method should be called after the `open()` method. For this reason I moved the line `LOG.info(writers.describe());` on `IndexingJob.java:124` to `IndexerOutputFormat.java:45`. This change is suitable because now the `IndexWriters.get(conf)` method isn't invoked twice.


r0ann3l:
Reviewing it again I see when the arguments of index or clean commands are not enough the `describe()` method is called before `open()` method. I think the use of method `describe()` here is not useful because:

1. Shows the default values of the attributes.
2. The error is due to there are not enough arguments, it is not a misconfigured index writer.

PR #360, opened 2018-07-11
NUTCH-2618 protocol-okhttp not to use http.timeout for max duration to fetch document

sebastian-nagel:
- add property http.time.limit to configure the max. time allowed to fetch a single document
- add reason of truncation (content or time) to response metadata
- rename "trimmed" -> "truncated" to follow common Nutch terminology

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -224,6 +224,20 @@
 </property>
 
 <property>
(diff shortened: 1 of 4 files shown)


sebastian-nagel:
Merged as part/precondition of #361

PR #363, opened 2018-07-17
NUTCH-2616 Review routing of deletions by Exchange component

sebastian-nagel:
- send deletions to all index writers as it was the case before the Exchange component has been introduced


r0ann3l:
+1 lgtm. Thanks @sebastian-nagel

PR #365, opened 2018-07-20
NUTCH-2621 Generate report of third-party licenses

sebastian-nagel:
- add ant target `ant report-licenses` (for core and plugins) which generates tabular reports of dependency licenses

The created table is comparable to those created by the [Maven license plugin](https://www.mojohaus.org/license-maven-plugin/index.html)
- it also shows dual licenses (important if one of the alternatives is compatible with the Apache license)
- must be called separately for core and each plugin, e.g.
```
# Nutch core
ant report-licenses
more build/org.apache.nutch-nutch-default-3rd-party-licenses.tsv

# plugin any23
cd src/plugin/any23
ant report-licenses
more ../../../build/any23/org.apache.nutch-any23-default-3rd-party-licenses.tsv
```

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -562,7 +562,13 @@
 
   <!-- target: report  ================================================== -->
   <target name="report" depends="resolve-test" description="--> generates a report of dependencies">
-    <ivy:report todir="${build.dir}"/>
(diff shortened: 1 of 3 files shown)


lewismc:
+1

PR #369, opened 2018-07-27
NUTCH-2623 Fetcher to guarantee delay for same host/domain/ip independent of http/https protocol

sebastian-nagel:
- the modes byHost, byDomain and byIP now do not include the protocol for assigning URLs to queues
- added fetcher.queue.mode `byHostProtocol` for legacy
- centralize check of queue mode


sebastian-nagel:
Updated and removed legacy mode `byHostProtocol` as discussed in NUTCH-2623.

PR #387, opened 2018-10-08
NUTCH-2630 Fetcher to log skipped records by robots.txt

sebastian-nagel:
Change required log level to INFO (default) for messages reporting skipped URLs because of robots.txt rules (disallow or crawl delay larger than fetcher.max.crawl.delay).


lewismc:
+1

PR #398, opened 2018-10-17
NUTCH-2658 Add README for the index-links plugin

jorgelbg:
Add a README file for the index-links plugin. At the very least, least this solves part of the issue with users knowing what they need to add to their backend (usually Solr).

### DIFF
--- a/src/plugin/index-links/README.md
+++ b/src/plugin/index-links/README.md
@@ -1,0 +1,53 @@
+indexer-links plugin for Nutch
+==============================
+
+This plugin provides the feature to index the inlinks and outlinks of a URL
+into an indexing backend.
+
+## Configuration
+
+This plugin provides the following configuration options:
+
+* `index.links.outlinks.host.ignore`: If true, the plugin will ignore outlinks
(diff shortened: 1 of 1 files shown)


src/plugin/index-links/README.md
```diff
@@ -0,0 +1,53 @@
...
+given URL will have multiple inlinks and outlinks.
+
+```
+<fieldType name="string" class="solr.StrField" sortMissingLast="true" omitNorms="true"/>
+```
+
+The field configuration could look like:
+
+```
+<field name="inlinks" type="multiValuedString" stored="true" indexed="true" multiValued="true"/>
```
sebastian-nagel: The Solr schema ([conf/schema.xml](/apache/nutch/blob/master/conf/schema.xml)) already contains the field definitions for multiple IndexingFilter plugins. Why not add inlinks and outlinks also to the schema?

jorgelbg: Yes, I realized last night that the fields are missing from the `conf/schema.xml` file. I'm going to add them there as well.

lewismc:
+1

PR #399, opened 2018-10-17
NUTCH-2661 Move the TestOutlinks class into the o.a.n.parse path

jorgelbg:
This test covers the specific case of the comparison betwen 2 identical `Outlink` instances. Because this is not `index-links` specific I'm moving the test class into the core parse tests.


sebastian-nagel:
+1 that's the right location

PR #402, opened 2018-10-23
Add the fields required by the index-links plugin to the Solr schema

jorgelbg:
As discussed in #398 it makes perfect sense to add the fields required by the index-links plugin to the `conf/schema.xml`.


lewismc:
+1

PR #404, opened 2018-10-24
NUTCH-2668 Integrate OWASP dependency checks as ant target

sebastian-nagel:
- add ant target "report-vulnerabilities" to generate report
- initial suppression list to exclude false positives

To install the tool (eventually the installation can be automatized as ant target):
```
wget http://dl.bintray.com/jeremy-long/owasp/dependency-check-ant-3.3.2-release.zip
unzip -d ivy/ dependency-check-ant-3.3.2-release.zip
```

Run it
```
ant report-vulnerabilities
```
and open the reports `build/dependency-check-report.html` or `build/dependency-check-vulnerability.html`. There are false positives, the tool allows to maintain an suppression list which needs to be completed.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -599,6 +599,35 @@
   </fail>
  </target>
 	
+  <!-- Check dependencies for security vulnerabilities                                    -->
+  <!-- requires installation of OWASP dependency check tool, see                          -->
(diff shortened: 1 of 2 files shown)


jorgelbg:
+1 Thanks @sebastian-nagel great job!

PR #428, opened 2019-01-16
NUTCH-2686 New property: "moreIndexingFilter.mapMimeTypes.field"

r0ann3l:
Includes a new property: "moreIndexingFilter.mapMimeTypes.field", which indicates the field's name where the mapped mime type must be written. If this property is NULL (default value) the field "type" will be replaced by the mapped mime type (current behavior).

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -1207,6 +1207,15 @@
   </description>
 </property>
 
+<property>
+  <name>moreIndexingFilter.mapMimeTypes.field</name>
+  <value></value>
+  <description>It's used if moreIndexingFilter.mapMimeTypes is true. Indicates the field
+  where the mapped MIME-type must be written. If it's null, the field "type"
+  will be replaced by the mapped MIME-type.
+  </description>
+</property>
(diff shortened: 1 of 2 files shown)


conf/nutch-default.xml
```diff
@@ -1207,6 +1207,15 @@
   </description>
 </property>
 
+<property>
+  <name>moreIndexingFilter.mapMimeTypes.field</name>
+  <value></value>
+  <description>It's used if moreIndexingFilter.mapMimeTypes is true. Indicates the field
+  where the mapped MIME-type must be written. If it's null, the field "type"
```
sebastian-nagel: I would prefer: `If it's empty or unset, the content of the field "type" ..."`

Of course, `conf.get("moreIndexingFilter.mapMimeTypes.field")` returns null in both cases (empty or unset). But that's only known to programmers familiar with the internals of a Hadoop configuration. Just to prevent that anybody writes literally "null" as value.

sebastian-nagel:
Thanks, @r0ann3l!

PR #460, opened 2019-08-09
NUTCH-2727 Upgrade Hadoop dependencies to 2.9.2

sebastian-nagel:
- fix unit tests of protocol-okhttp to use plugin class loader
  and methods defined by Protocol interface to avoid dependency
  conflicts (Hadoop depends on okhttp as well)


sebastian-nagel:
Rebased to master, committing now...

PR #465, opened 2019-08-27
NUTCH-2719: An exchange points to an index writer not available

r0ann3l:
When an exchange points to a non-existent index writer, Nutch logs a warning indicating the inconsistent configuration and avoids the current NPE.


sebastian-nagel:
+1

PR #466, opened 2019-08-28
NUTCH-2718: Filenames of configuration files of index writers and exchanges

r0ann3l:
With this patch, select another file for exchanges and index writers configuration is available out-of-the-box. For this purpose there are two new properties: `exchanges.exchanges.file` and `indexer.indexwriters.file`


sebastian-nagel:
+1 lgtm.

PR #468, opened 2019-09-01
NUTCH-2654: The obsolete configuration of index writers has been removed.

r0ann3l:
In this patch, all obsolete files and properties associated with indexers have been removed.


r0ann3l:
In this PR the schema.xml file has not been deleted. I assume that this file is kept in Nutch distribution as reference of how Solr has to be configured. But I believe this file has to be removed as well, because:
- Maybe this file was necessary when Nutch had only the Solr index writer out-of-the-box. But now, there are more index writers available in Nutch distribution. So, why do we have to keep this file? Why we don't have references of the configuration of other indexers available in Nutch too?
- In the mail list I've seen some people a little confuse about the use of this file. Sometimes the users think that this file has to match with the schema.xml file in Solr side, when actually this file is never used by Nutch at all.
So, what do you think about it folks?

jorgelbg:
+1 for removing the outdated config from the `nutch-default.xml`.

I'm not so sure about the removal of the `schema.xml` file. Looking at our supported indexer must of them don't need a pre-defined schema (Kafka, rabbitmq, elastic-rest, elastic, etc.). For Solr, a schema is normally required (although recent versions do have a [schemaless mode](https://lucene.apache.org/solr/guide/7_0/schemaless-mode.html#using-the-schemaless-example).  The issue boils down to how the user starts Solr.

If a Solr instance was started without schemaless mode then it will fail to index the data in which case there is no clear way of knowing which fields to add in the schema other than trial&error (or checking each enabled plugin)

If the user starts Solr with the schemaless mode, the schema is not entirely needed.

I do recognize that this file is used by a single indexer plugin which is a conditional dependency (using or not Solr). We _could_ move the config file into the plugin directory (`src/plugin/indexer-solr`) and update the documentation/README making the file available on the distribution (otherwise the user will have to download it). It is possible that putting it in this new location makes it clear that it is not used by Nutch itself but provided for the Solr users.

The `schema.xml` file is not required but provided as a commodity/guide for Solr users.

sebastian-nagel:
+1

The schema.xml but it isn't removed by this PR, right? It should be kept, to make the setup of Solr easy for users. If I'm not wrong, the solrindex-mapping.xml is obsolete because a field mapping can be configured in the index-writers.xml.

r0ann3l:
I agree with @jorgelbg. We could move `schema.xml` from folder` conf/` to` src/plugin/indexer-solr`. After all, this is not a configuration file of Nutch. It is only a reference/guide of how to configure a Solr core.

sebastian-nagel:
+1 Agreed to move the the schema.xml to the indexer-solr plugin. We just need to update the tutorial regarding this point as soon as 1.16 is released. Might be also worth a note in the CHANGES.txt

PR #469, opened 2019-09-05
NUTCH-2732: .template for configuration files.

r0ann3l:
In order to avoid ignored and tracked files by git at the same time I renamed those files to extension `.template`.


sebastian-nagel:
+1

PR #470, opened 2019-09-09
NUTCH-2734 Upgrade Tika dependency to 1.22

sebastian-nagel:


jorgelbg:
+1 LGTM

@sebastian-nagel Thanks for taking care of this. I was using Tika in a different project and ran into [TIKA-1568](https://issues.apache.org/jira/browse/TIKA-1568). Since Nutch was using an older version as well I wanted to do this as well.

PR #478, opened 2019-09-30
NUTCH-2279 LinkRank fails when using Hadoop MR output compression

sebastian-nagel:
- read output directory of link counter job to determine output file name (fail if there is none or more than one file)
- determine output codec and use it to read the output


lewismc:
I've never come across this issue. What is the error trace like when this is encountered? Is a test required?

sebastian-nagel:
@naegelejd added the stack trace in Jira. A test isn't really necessary, just tried it in local mode with mapreduce.output.fileoutputformat.compress enabled or disabled. Will also test later in pseudo-distributed mode when testing the release candidate.

PR #479, opened 2019-10-10
Nutch-2511 Support large sitemaps

YossiTamari:
Override http.content.limit with sitemap.size.max when fetching sitemaps. Defaults to 50MB.

### DIFF
--- a/src/java/org/apache/nutch/util/SitemapProcessor.java
+++ b/src/java/org/apache/nutch/util/SitemapProcessor.java
@@ -108,7 +108,9 @@ public class SitemapProcessor extends Configured implements Tool {
 
     public void setup(Context context) {
       Configuration conf = context.getConfiguration();
-      conf.setInt("http.content.limit", conf.getInt(SITEMAP_SIZE_MAX, SiteMapParser.MAX_BYTES_ALLOWED));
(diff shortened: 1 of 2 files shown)


src/java/org/apache/nutch/util/SitemapProcessor.java
```diff
@@ -107,6 +108,7 @@
 
     public void setup(Context context) {
       Configuration conf = context.getConfiguration();
+      conf.setInt("http.content.limit", conf.getInt(SITEMAP_SIZE_MAX, SiteMapParser.MAX_BYTES_ALLOWED));
```
sebastian-nagel: Also set `file.content.limit`? - just to make sure that a local sitemap also works and if it only has been downloaded for testing.

Please add the new property to nutch-default.xml, there is already a "section" listing "sitemap properties". Agreed that users rarely will adjust it. However, the nutch-default.xml is the single place where all properties are documented. Should be part of our documentation, see [NUTCH-2743](https://issues.apache.org/jira/browse/NUTCH-2743).

YossiTamari: Personally I think we should separate documentation from the actual configuration file, but I made both requested changes.

PR #481, opened 2019-10-17
NUTCH-1559 parse-metatags duplicates extracted metatags

sebastian-nagel:
- do not add metatags already in ParseData's as this may lead to duplicates
- add unit test
- fix logging in MetaTagsParser to use slf4j

### DIFF
--- a/src/plugin/parse-metatags/src/java/org/apache/nutch/parse/metatags/MetaTagsParser.java
+++ b/src/plugin/parse-metatags/src/java/org/apache/nutch/parse/metatags/MetaTagsParser.java
@@ -23,8 +23,6 @@ import java.util.Locale;
 import java.util.Properties;
 import java.util.Set;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.nutch.metadata.Metadata;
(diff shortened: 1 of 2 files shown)


src/plugin/parse-metatags/src/java/org/apache/nutch/parse/metatags/MetaTagsParser.java
```diff
@@ -16,21 +16,22 @@
...
 package org.apache.nutch.parse.metatags;
 
+import java.lang.invoke.MethodHandles;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Properties;
 import java.util.Set;
 
-import org.apache.commons.logging.Log;
```
lewismc: I wonder why this is even on the cp. Can we remove it?

lewismc: Classpath

sebastian-nagel: It's a transitive dependency of commons-jexl (and also of few libs required for tests). There are still a few Nutch classes using it instead of org.slf4j. I've opened [NUTCH-2747](https://issues.apache.org/jira/browse/NUTCH-2747) to address this.

PR #483, opened 2019-10-24
Fix for NUTCH-2750

sju:


lewismc:
@sju how would one go about testing this patch? Thank you

sju:
@lewismc We run it like this:
`bin/nutch readdb <crawldb> -listen <port>`
It will then listen on the provided port for URL's.

sebastian-nagel:
+1 lgtm. I'll try to test it in the wild. Thanks, @sju!

sebastian-nagel:
Hi @sju,

I've verified that the readers are refreshed if the CrawlDb has changed - tested in local mode on a local file and in distributed mode with a CrawlDb on HDFS.

Will commit soon. Thanks!

Actually, I'm really impressed how fast look-ups are once the readers are initialized: I've got 20 look-ups per second on a CrawlDb containing 1 billion records in 200 partitions hold on HDFS. That's pretty fast given that it's not a "hot" index but only files on HDFS. Opened [NUTCH-2753](https://issues.apache.org/jira/browse/NUTCH-2753) to add this great option to the command-line help of readdb/readlinkdb.

PR #490, opened 2019-12-21
Fix for NUTCH-1863: Add JSON format dump output to readdb command

balashashanka:
Added JSON dump capability to readdb command line tool.

### DIFF
--- a/src/java/org/apache/nutch/crawl/CrawlDbReader.java
+++ b/src/java/org/apache/nutch/crawl/CrawlDbReader.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -75,6 +76,12 @@ import org.apache.nutch.util.SegmentReaderUtil;
 import org.apache.nutch.util.StringUtil;
 import org.apache.nutch.util.TimingUtil;
 import org.apache.commons.jexl2.Expression;
+
+import com.fasterxml.jackson.core.JsonGenerationException;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.util.MinimalPrettyPrinter;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 
 /**
  * Read utility for the CrawlDB.
@@ -128,6 +135,26 @@ public class CrawlDbReader extends AbstractChecker implements Closeable {
       }
     }
     readers = null;
+  }
+
+  public static class JsonIndenter extends MinimalPrettyPrinter {
+
+    /**
+    * 
+    */
+    private static final long serialVersionUID = -4464852619186879060L;
+
+    // @Override
+    public void writeObjectFieldValueSeparator(JsonGenerator jg)
+        throws IOException, JsonGenerationException {
+      jg.writeRaw(": ");
+    }
+
+    // @Override
+    public void writeObjectEntrySeparator(JsonGenerator jg)
+        throws IOException, JsonGenerationException {
+      jg.writeRaw(", ");
+    }
   }
 
   public static class CrawlDatumCsvOutputFormat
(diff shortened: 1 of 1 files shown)


src/java/org/apache/nutch/crawl/CrawlDbReader.java
```diff
@@ -185,13 +190,83 @@ public synchronized void write(Text key, CrawlDatum value)
...
+      return new LineRecordWriter(fileOut);
+    }
+  }
+
+  public static class CrawlDatumJsonOutputFormat
+      extends FileOutputFormat<Text, CrawlDatum> {
+    protected static class LineRecordWriter
+        extends RecordWriter<Text, CrawlDatum> {
+      private DataOutputStream out;
+      private ArrayList<String> jsonString = new ArrayList<String>();
```
sebastian-nagel: Nutch is designed to be scalable and should be able to handle CrawlDbs containing billions of items. Even, if the CrawlDb is split into many partitions, every task has to handle 100 million CrawlDb items. However, it's hardly possible to keep such a huge list in memory, resp. it would be a too expensive use of resources. So, it's mandatory to write each record directly to the output, Hadoop itself handles the large outputs efficiently by buffering parts in memory and spilling to disk if the buffers are full.

To simplify the process: why not use the [JSON lines format](http://jsonlines.org/)? Every line one JSON-formatted record: `{"url": "...", ...}`
- makes it much easier to write the output because you need not to preserver state internally to properly open and close the list of records.
- JSON lines has several other advantages:
  - it's possible to grep for records
  - also [jq](https://stedolan.github.io/jq/) can process it
  - and again: any JSON parser needs to keep only a single line in memory

src/java/org/apache/nutch/crawl/CrawlDbReader.java
```diff
@@ -185,13 +190,83 @@ public synchronized void write(Text key, CrawlDatum value)
...
+        } catch (IOException e) {
+        }
+      }
+
+      public synchronized void write(Text key, CrawlDatum value)
+          throws IOException {
+        String fetchTime = new Date(value.getFetchTime()).toString();
+        String modifiedTime = new Date(value.getModifiedTime()).toString();
+        String recordString = "";
+        recordString += "{\n" + "\t\"url\":\"" + key.toString()
```
sebastian-nagel: I would strongly recommend to use a JSON library to do the escaping. There are many pitfalls, e.g., what if a URL contains a `"`? [Jackson](https://github.com/FasterXML/jackson) is already on board (it's a dependency) and it's pretty handy to write JSON:
```java
import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.ObjectWriter;
...
public static class JsonIndenter extends MinimalPrettyPrinter {

    // @Override
    public void writeObjectFieldValueSeparator(JsonGenerator jg)
        throws IOException, JsonGenerationException {
      jg.writeRaw(": ");
    }

    // @Override
    public void writeObjectEntrySeparator(JsonGenerator jg)
        throws IOException, JsonGenerationException {
      jg.writeRaw(", ");
    }
}
...
// inside write(Text key, CrawlDatum value)
ObjectMapper jsonMapper = new ObjectMapper();
jsonMapper.getFactory().configure(JsonGenerator.Feature.ESCAPE_NON_ASCII,
        true);
ObjectWriter jsonWriter = jsonMapper.writer(new JsonIndenter());

...
Map<String, String> data = new LinkedHashMap<String, String>(); // a linked hash map preserves the ordering
data.put("url", key.toString());
// put all other fields into the map
...
// and write the serialized record
out.write(jsonWriter.writeValueAsBytes(data));
```

There are many options to control the output format...

balashashanka:
Hi @sebastian-nagel, I have made the changes. Thanks for the insight on the reason for using DataOutputStream. In the write function I have used Map<String, Object> since the metadata is a json by itself.

src/java/org/apache/nutch/crawl/CrawlDbReader.java
```diff
@@ -879,35 +988,37 @@ public void processTopNJob(String crawlDb, long topN, float min,
...
           .println("\t<crawldb>\tdirectory name where crawldb is located");
       System.err
           .println("\t-stats [-sort] \tprint overall statistics to System.out");
       System.err.println("\t\t[-sort]\tlist status sorted by host");
-      System.err
-          .println("\t-dump <out_dir> [-format normal|csv|crawldb]\tdump the whole db to a text file in <out_dir>");
+      System.err.println(
+          "\t-dump <out_dir> [-format normal|csv|crawldb|json]\tdump the whole db to a text file in <out_dir>");
       System.err.println("\t\t[-format csv]\tdump in Csv format");
-      System.err
```
sebastian-nagel: Should also add a line describing `-format json`, maybe clarify that it's "JSON lines".

sebastian-nagel:
Thanks, @balashashanka! Look great!

PR #508, opened 2020-04-11
NUTCH-2757 : Indexer-elastic: add authentication options

balashashanka:
Added username and password as default credentials to the HTTP Client

### DIFF
--- a/conf/index-writers.xml.template
+++ b/conf/index-writers.xml.template
@@ -108,8 +108,10 @@
     <parameters>
       <param name="host" value="localhost"/>
       <param name="port" value="9200"/>
-      <param name="cluster" value=""/>
       <param name="index" value="nutch"/>
+      <!--Configure username and password if necessary, default user is "elastic"
+      <param name="user" value="user"/>
+      <param name="password" value="password"/>-->
       <param name="max.bulk.docs" value="250"/>
       <param name="max.bulk.size" value="2500500"/>
       <param name="exponential.backoff.millis" value="100"/>
--- a/conf/index-writers.xml.template
+++ b/conf/index-writers.xml.template
@@ -109,9 +109,9 @@
       <param name="host" value="localhost"/>
       <param name="port" value="9200"/>
       <param name="index" value="nutch"/>
-      <!--Configure username and password if necessary, default user is "elastic"
-      <param name="user" value="user"/>
-      <param name="password" value="password"/>-->
+      <param name="username" value="elastic"/>
+      <param name="password" value=""/>
+      <!--<param name="auth" value="false"/>-->
       <param name="max.bulk.docs" value="250"/>
       <param name="max.bulk.size" value="2500500"/>
       <param name="exponential.backoff.millis" value="100"/>
--- a/src/plugin/indexer-elastic/README.md
+++ b/src/plugin/indexer-elastic/README.md
@@ -32,8 +32,9 @@ Parameter Name | Description | Default value
 --|--|--
 host | Comma-separated list of hostnames to send documents to using [TransportClient](https://static.javadoc.io/org.elasticsearch/elasticsearch/5.3.0/org/elasticsearch/client/transport/TransportClient.html). Either host and port must be defined or cluster. | 
 port | The port to connect to using [TransportClient](https://static.javadoc.io/org.elasticsearch/elasticsearch/5.3.0/org/elasticsearch/client/transport/TransportClient.html). | 9300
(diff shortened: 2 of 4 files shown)


src/plugin/indexer-elastic/src/java/org/apache/nutch/indexwriter/elastic/ElasticIndexWriter.java
```diff
@@ -162,6 +191,17 @@ protected RestHighLevelClient makeClient(IndexWriterParams parameters) throws IO
       } else	{
         LOG.debug("No cluster name provided so using default");
       }
+      //Only add username and password if password is configured
+      if(StringUtils.isNotBlank(password)) {
```
sebastian-nagel, line 195: Great, @balashashanka! Looks good and should work.

Minor point: If possible try to apply the [Nutch Eclipse Code Formatting rules](https://github.com/apache/nutch/blob/master/eclipse-codeformat.xml) and also remove trailing white space. Are you using Eclipse? If not let me know or whether you need help to set it up.

balashashanka: Hi @sebastian-nagel, did the formatting.

conf/index-writers.xml.template
```diff
@@ -108,8 +108,10 @@
     <parameters>
       <param name="host" value="localhost"/>
       <param name="port" value="9200"/>
-      <param name="cluster" value=""/>
       <param name="index" value="nutch"/>
+      <!--Configure username and password if necessary, default user is "elastic"
+      <param name="user" value="user"/>
```
jorgelbg, line 113: Nit: Perhaps we should use `username` instead of `user`. For the Solr indexer configuration, we use `username` and also the comment mentions `username`.

src/plugin/indexer-elastic/src/java/org/apache/nutch/indexwriter/elastic/ElasticIndexWriter.java
```diff
@@ -114,57 +136,78 @@ public void open(IndexWriterParams parameters) throws IOException {
...
-        Header[] defaultHeaders = new Header[]{new BasicHeader("cluster.name", cluster)};
+        Header[] defaultHeaders = new Header[] {
+            new BasicHeader("cluster.name", cluster) };
         restClientBuilder.setDefaultHeaders(defaultHeaders);
-      } else	{
+      } else {
         LOG.debug("No cluster name provided so using default");
       }
+      // Only add username and password if password is configured
+      if (StringUtils.isNotBlank(password)) {
```
jorgelbg, line 197: On the Solr indexer, we have a dedicated `auth` `boolean` key to verify if the authentication must be set (see https://github.com/apache/nutch/blob/master/src/plugin/indexer-solr/src/java/org/apache/nutch/indexwriter/solr/SolrIndexWriter.java#L97). I'm not sure I have a *strong* preference for either option (the boolean `auth` key vs checking for the empty `password` key) but I would err in the side of consistency since both Indexers work in a similar way.

balashashanka: Hi @jorgelbg, thanks for pointing this out. I can change the code to follow the same.

jorgelbg: @balashashanka sure, let's make this part the same for both indexers

conf/index-writers.xml.template
jorgelbg: Nit: perhaps here we should use as well a default `elastic` value for the username?

src/plugin/indexer-elastic/README.md
```diff
@@ -32,10 +32,12 @@ Parameter Name | Description | Default value
 --|--|--
 host | Comma-separated list of hostnames to send documents to using [TransportClient](https://static.javadoc.io/org.elasticsearch/elasticsearch/5.3.0/org/elasticsearch/client/transport/TransportClient.html). Either host and port must be defined or cluster. |
```
jorgelbg, line 33: I missed that the `or cluster` part of the sentence is now obsolete (due to the removal) but that is not a blocker from my side. If you fix it before merging it would be great.

balashashanka: Hi, kind of missed this one. I removed this in the latest commit.

PR #512, opened 2020-04-23
NUTCH-2781 Increase default Java heap size

sebastian-nagel:
- increase default value for NUTCH_HEAPSIZE to 4096 MB (from 1000 MB)
- remove -Dmapred.child.java.opts=-Xmx1000m from default options in bin/crawl

PR #513, opened 2020-04-23
NUTCH-2501 allow to set Java heap size when using crawl script in distributed mode

sebastian-nagel:
- bin/crawl
   - add hint how to set map and reduce task memory via -D ... options
   - use -D options for all steps (Nutch tools), fixes [NUTCH-2379](https://issues.apache.org/jira/projects/NUTCH/issues/NUTCH-2379)
   - fix quoting of -D options, eg. -D plugin.includes='protocol-xyz|parse-xyz'
   - use -D options for all steps (Nutch tools)
- bin/nutch
  - document that environment variables are only used in local mode

(includes #512 / NUTCH-2781)

### DIFF
--- a/src/bin/crawl
+++ b/src/bin/crawl
@@ -23,7 +23,13 @@
 #
 # Options:
 #   -i|--index                            Indexes crawl results into a configured indexer
-#   -D                                    A Java property to pass to Nutch calls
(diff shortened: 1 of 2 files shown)


sebastian-nagel:
Rebased to master.

PR #514, opened 2020-04-23
NUTCH-1194 Generator: CrawlDB lock should be released earlier

sebastian-nagel:
- release CrawlDb lock after select step, in case, generated items are not marked in CrawlDb (generate.update.crawldb is false)


sebastian-nagel:
Rebased to master and squashed commits.

PR #532, opened 2020-06-09
NUTCH-2790 indexer-csv: escape field leading quote character

pmezard:
Before the change, the leading quote of a field value like '"value'
would be left unescaped.

NOTE: the tests fail locally with and without the PR. The only output I have is:
```
test:
     [echo] Testing plugin: urlnormalizer-slash
    [junit] Running org.apache.nutch.net.urlnormalizer.regex.TestRegexURLNormalizer
    [junit] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.276 sec
    [junit] Running org.apache.nutch.net.urlnormalizer.slash.TestSlashURLNormalizer
    [junit] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.306 sec
```
Is there any way to have detailed test information when running them from command line? I am not very excited about working with Eclipse.

### DIFF
--- a/src/plugin/indexer-csv/src/java/org/apache/nutch/indexwriter/csv/CSVIndexWriter.java
+++ b/src/plugin/indexer-csv/src/java/org/apache/nutch/indexwriter/csv/CSVIndexWriter.java
@@ -405,13 +405,12 @@ public class CSVIndexWriter implements IndexWriter {
     if (max > maxFieldLength) {
       max = maxFieldLength;
     }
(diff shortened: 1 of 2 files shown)


sebastian-nagel:
+1 Good catch!

> Is there any way to have detailed test information when running them from command line?

You can execute the unit tests for a specific plugin:
```
ant test-plugin -Dplugin=indexer-csv
```
The test output is written to `build/indexer-csv/test/`(output of core tests is in `build/test`).

PR #537, opened 2020-07-10
[NUTCH-2801] RobotsRulesParser command-line checker to use http.robots.agents as fall-back

sebastian-nagel:
- if no agent names are given as command-line arguments use values ofhttp.agent.name and http.robots.agents as agent names to be checked
- update command-line help

```
$> nutch org.apache.nutch.protocol.RobotRulesParser \
      -Dhttp.agent.name='mybot' \
      -Dhttp.robots.agents='nutch,goodbot' \
      robots.txt urls.txt 
Testing robots.txt for agent names: mybot,nutch,goodbot
allowed:        https://www.example.com/

# command-line overwrite:
$> nutch org.apache.nutch.protocol.RobotRulesParser \
      -Dhttp.agent.name='mybot' \
      -Dhttp.robots.agents='nutch,goodbot' \
      robots.txt urls.txt \
      badbot,anybot
Testing robots.txt for agent names: badbot,anybot
not allowed:    https://www.example.com/
```

### DIFF
--- a/src/java/org/apache/nutch/protocol/RobotRulesParser.java
+++ b/src/java/org/apache/nutch/protocol/RobotRulesParser.java
@@ -133,7 +133,7 @@ public abstract class RobotRulesParser implements Tool {
 
     String[] confWhiteList = conf.getStrings("http.robot.rules.whitelist");
     if (confWhiteList == null) {
-      LOG.info("robots.txt whitelist not configured.");
+      LOG.debug("robots.txt whitelist not configured.");
     }
     else {
       for (int i = 0; i < confWhiteList.length; i++) {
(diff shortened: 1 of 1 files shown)


src/java/org/apache/nutch/protocol/RobotRulesParser.java
```diff
@@ -376,13 +379,18 @@ public int run(String[] args) {
    */
   private static class TestRobotRulesParser extends RobotRulesParser {
 
-    public TestRobotRulesParser(Configuration conf) {
+    public void setConf(Configuration conf) {
       // make sure that agent name is set so that setConf() does not complain,
```
balashashanka, line 383: Hi @sebastian-nagel, a small question. Is this comment still valid, since if the user-agent is not provided we will check with robots.agent and others right?

sebastian-nagel: Thanks. You're right the comment wasn't up-to-date. Would have been simpler to drop the command-line overwrite of the checked agent names and rely only on properties. But I didn't want to break the existing behavior.

PR #539, opened 2020-07-10
NUTCH-2803 Rename property http.robot.rules.whitelist

lewismc:
This issue addresses https://issues.apache.org/jira/browse/NUTCH-2803

Feels good to be contributing to Nutch again ... !!!

Use of language here... **white** being replaced with **allow**


conf/nutch-default.xml
```diff
@@ -118,7 +118,7 @@
 </property>
 
 <property>
-  <name>http.robot.rules.whitelist</name>
+  <name>http.robot.rules.allowlist</name>
```
sebastian-nagel, line 121: Just thinking about the naming of the property and whether `http.robotstxt.ignore` might be the better naming. Or is there is a more precise alternative?

Let's first state what the property effects: If a host or domain is listed in the property
- no robots.txt is fetched and an empty robots rules set is assumed for this host. Consequently, for this host:
- all URLs are allowed (no URL is disallowed)
- the Crawl-Delay is ignored
- no sitemaps are detected

In short, it has many more effects than "allow" any URL from the specified host. Also, robots rules in HTML metadata or HTTP headers (noindex, nofollow) are not affected and may still cause that pages from the specified host are not indexed.

Also important to emphasize: robots.txt rules tell search engine crawlers what is (not) worth to index: exclude duplicates (same content with different view for print, re-sorted, re-formatted), skip over undesired content (private), etc.  That's why it might be smarter to [follow the rules of a more privileged bot](http://www.thesempost.com/apples-applebot-follows-googlebots-instructions-in-robots-txt-files/).

All these disadvantages and site-effects of ignoring the robots.txt should be documented, see [NUTCH-2808](https://issues.apache.org/jira/browse/NUTCH-2808) and [NUTCH-2807](https://issues.apache.org/jira/browse/NUTCH-2807).

sebastian-nagel: Hi @lewismc (or anybody else), any preferences how to name the property or any suggestions for alternative solutions?

balashashanka: Hi @sebastian-nagel, I think that "allowlist" gives good context from user perspective. But since we still have to document and clearly explain the disadvantages. We can go ahead with http.robots.txt.ignore (or similar).

sebastian-nagel:
Resolved the merge conflicts - no wonder after over one year that this PR is sitting around. Waiting for the CI build, will then commit and later address NUTCH-2808 and NUTCH-2807 in a separate PR. Thanks, @lewismc and @balashashanka!

PR #542, opened 2020-07-27
NUTCH-2810 FreeGenerator to actually apply configured number of fetch lists

sebastian-nagel:


sebastian-nagel:
Successfully tested in pseudo-distributed using Hadoop 3.3.0: `-numFetchers <n>` is correctly applied.

PR #543, opened 2020-07-31
NUTCH-2811 : Setup Github workflows for prs

madhawa-gunasekara:
Setup Github workflows

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -1,0 +1,36 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: master pr build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ '1.8' ]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Build with Ant
+        run: ant -noinput -buildfile build.xml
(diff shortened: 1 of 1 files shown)


.github/workflows/master-build.yml
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: master pr build
+
+on: [push, pull_request]
```
balashashanka, line 20: Hi @madhawa-gunasekara, the check runs well. But When i tried it in own repo [here ](https://github.com/balashashanka/nutch/tree/NUTCH-2181) and [here](https://github.com/balashashanka/nutch/pull/1). For a PR same check is ran twice. So may be run this only for PR or split this to run check on master after merge and for a PR.

.github/workflows/master-build.yml
```diff
@@ -0,0 +1,36 @@
...
+        java: [ '1.8' ]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Build with Ant
+        run: ant clean nightly javadoc -buildfile build.xml
```
balashashanka, line 36: Do we need to generate javadoc here? Or just nightly will do?

.github/workflows/master-build.yml
madhawa-gunasekara: Hi @balashashanka 
Here you can see, I added workflows for both pull requests and direct pushes. so basically It will run 1 time for the pull request, and one time after the merge. (push). I haven't mentioned a specific branch for pushes. so this will enable for every branch. [In Gora](https://github.com/apache/gora/blob/master/.github/workflows/master-push-build.yml#L22), we have specified master branch for pushes. Let me know if we need to specify particular branches for pushes.

.github/workflows/master-build.yml
madhawa-gunasekara: I have taken these parameters from the [Jenkins configurations.](https://builds.apache.org/view/M-R/view/Nutch/job/Nutch-trunk/configure)

.github/workflows/master-build.yml
balashashanka: Yes it would be nice if we add it to master as it will be triggered only once for every pull request.

madhawa-gunasekara: Don't you need for 2.x branch?

balashashanka: Since we dont have an official dev support on 2.x I dont think we would need it. 
But @sebastian-nagel or @lewismc can give a bit of clarity on this.
Or in the mean time we can create a different PR with file for 2.x and merge if its required.

jorgelbg: Agreed, 2.x branch development was stopped last year and I don't think we're going to accept any contributions to that branch on the foreseeable future.

madhawa-gunasekara: Okay, I updated

lewismc:
Nice @madhawa-gunasekara good work

balashashanka:
LGTM, thanks @madhawa-gunasekara

PR #553, opened 2020-10-11
NUTCH-2809 Upgrade any23 plugin dependency to 2.4

lewismc:
This issue addresses https://issues.apache.org/jira/browse/NUTCH-2809

All ClassPath issues are resolved. 

I think we may be able to improve usability of this plugin if we included more extractors in the [any23.extractors](https://github.com/apache/nutch/blob/master/conf/nutch-default.xml#L1263-L1267) property.

Any comments/suggestions?


lewismc:
@balashashanka FYI :)

lewismc:
I'll note that the unit tests had to be adapted to accommodate an increase in the number of extractions performed by Any23 extractors.

sebastian-nagel (approved):
Thanks, @lewismc! Unit tests pass and I've also tested some Any23 extractors using parsechecker.

PR #566, opened 2021-01-28
NUTCH-2846 Fix various bugs spotted by Spotbugs

sebastian-nagel:
- Remove invalid declaration of constructor
- Use integer arithmetic instead of floating point with rounding
- Fix incorrect bracketing in calculation of hash code
- Parse integer numbers via Integer.parseInt(...)


lewismc:
Fantastic effort @sebastian-nagel . I had a feeling you were busy at work with something ;)

lewismc:
+1

PR #570, opened 2021-02-18
NUTCH-2850 Method ignores exceptional return value

lewismc:
PR to address https://issues.apache.org/jira/browse/NUTCH-2850
I've confirmed that Spotbugs no longer flags this issue.


sebastian-nagel (approved):
+1

PR #571, opened 2021-02-18
NUTCH-2851 Random object created and used only once

lewismc:
This PR addresses https://issues.apache.org/jira/browse/NUTCH-2851
I chose the following proposed fix...

Create a constant Random in the offending class
```
private static final Random RANDOM = new Random();
```
We then reuse as follows
```
RANDOM.nextInt();
```
This way, no new instances are created.


jorgelbg (approved):
+1 LGTM

sebastian-nagel:
+1

PR #647, opened 2021-05-20
NUTCH-2864 Upgrade Dockerfile to use JDK 11

lewismc:
This PR addresses https://issues.apache.org/jira/browse/NUTCH-2864
Using the Alpine Linux base is supposed to create a smaller overall image footprint...

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,18 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:18.04
+FROM alpine:3.13
 MAINTAINER Apache Nutch Committers <dev@nutch.apache.org>
 
 WORKDIR /root/
 
 
 # Install dependencies
-RUN apt update
-RUN apt install -y ant git openjdk-8-jdk-headless
-
-# Set up JAVA_HOME
-RUN echo 'export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64' >> $HOME/.bashrc
+RUN apk update
+RUN apk --no-cache add apache-ant git openjdk11
 
 # Checkout and build the Nutch master branch (1.x)
 RUN git clone https://github.com/apache/nutch.git nutch_source && cd nutch_source && ant runtime
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,13 +18,16 @@ MAINTAINER Apache Nutch Committers <dev@nutch.apache.org>
 
 WORKDIR /root/
 
-
 # Install dependencies
 RUN apk update
-RUN apk --no-cache add apache-ant git openjdk11
+RUN apk --no-cache add apache-ant bash git openjdk11
 
 # Checkout and build the Nutch master branch (1.x)
-RUN git clone https://github.com/apache/nutch.git nutch_source && cd nutch_source && ant runtime
+RUN git clone https://github.com/apache/nutch.git nutch_source && \
+     cd nutch_source && \
+     ant runtime && \
+     rm -rf build/ && \
+     rm -rf /root/.ivy2/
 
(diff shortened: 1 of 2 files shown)


lewismc:
Some details about the image
```
docker % docker build -t apache/nutch .
```
**TOTAL BUILD TIME**: 799.3s (overwhelming majority actually building Nutch)
**TOTAL SIZE**:  2.41GB (as opposed to 2.66GB from ubuntu 18 base image 7df7c861996f   6 weeks ago     2.66GB)

lewismc:
Further details
[ubuntu:18.04](https://hub.docker.com/layers/ubuntu/library/ubuntu/18.04/images/sha256-ceed028aae0eac7db9dd33bd89c14d5a9991d73443b0de24ba0db250f47491d2?context=explore) compressed size 25.46MB
[alpine:3.13](https://hub.docker.com/layers/alpine/library/alpine/3.13/images/sha256-9663906b1c3bf891618ebcac857961531357525b25493ef717bca0f86f581ad6?context=explore) compressed size 2.31 MB

lewismc:
We are talking ~250MB reduction in size. Which could be significant depending on the use case.

sebastian-nagel (changes requested):
Thanks @lewismc!

Alpine ships with busybox which integrates the Almquist shell (symlinked /bin/sh and /bin/ash): bin/nutch and bin/crawl require the bash in the shebang-line and also use bash-specific syntax
- esp. arrays are not available when using ash or any plain POSIX shell. Without arrays allowing white space in path names is likely to cause some headaches. Might be easier to install the bash instead of trying to make the shell scripts POSIX-compatible.
- otherwise would need to update docker/README.md and replace /bin/bash by /bin/sh

If we care about the size of the resulting image (we should, imho), maybe remove the ivy cache and build folders (maybe also runtime/deploy):
```
$> docker run --rm -ti  apache/nutch /bin/sh -c 'du -hs ~/nutch_source/build ~/.ivy2/ ~/nutch_source/runtime/deploy'
732.5M  /root/nutch_source/build
381.6M  /root/.ivy2/
331.2M  /root/nutch_source/runtime/deploy
```

lewismc:
@sebastian-nagel thanks for reviewing. Can you specifically let me know what actions you want me to take?

lewismc:
Thanks @sebastian-nagel I've updated and added some badges which give the consumer more details about the image.

sebastian-nagel (approved):

sebastian-nagel:
+1

docker/Dockerfile
```diff
@@ -13,21 +13,21 @@
...
 
 WORKDIR /root/
 
-
 # Install dependencies
-RUN apt update
-RUN apt install -y ant git openjdk-8-jdk-headless
-
-# Set up JAVA_HOME
-RUN echo 'export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64' >> $HOME/.bashrc
```
sebastian-nagel, line 27: JAVA_HOME needs to be set
```
RUN echo 'export JAVA_HOME=/usr/lib/jvm/java-11-openjdk' >> $HOME/.bashrc
```
otherwise bin/nutch fails:
```
bash-5.1# bin/nutch parsechecker https://nutch.apache.org/
Error: JAVA_HOME is not set.
```

sebastian-nagel (changes requested):
JAVA_HOME needs to be set:
/usr/lib/jvm/java-11-openjdk
otherwise bin/nutch fails:
```
bash-5.1# bin/nutch parsechecker https://nutch.apache.org/
Error: JAVA_HOME is not set.
```

lewismc:
I addressed the `JAVA_HOME` issue thanks @sebastian-nagel

sebastian-nagel (approved):
+1

PR #681, opened 2021-06-15
fireant upgrade dependency httpcore in ivy/ivy.xml from 4.4.9 to 4.4.14

lewismc:
fireant upgrade dependency httpcore in ivy/ivy.xml from 4.4.9 to 4.4.14


lewismc:
Hi folks... been open for 10 days now, I am merging. We can revert if we find problems down the road.

PR #689, opened 2021-06-25
NUTCH-2881 bug in 'nutch' symlink in docker container

lewismc:
Addresses https://issues.apache.org/jira/browse/NUTCH-2881

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,4 +32,4 @@ RUN git clone https://github.com/apache/nutch.git nutch_source && \
      rm -rf /root/.ivy2/
 
 # Convenience symlink to Nutch runtime local
-RUN ln -s nutch_source/runtime/local $HOME/nutch
+RUN ln -s nutch_source/runtime/local/bin/nutch $HOME/nutch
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -32,4 +32,4 @@ RUN git clone https://github.com/apache/nutch.git nutch_source && \
      rm -rf /root/.ivy2/
 
 # Convenience symlink to Nutch runtime local
```
sebastian-nagel, line 34: The comment should reflect that the symlink now points to the script bin/nutch.

Just two ideas:
- defining the environment variable NUTCH_HOME could be useful:
  ```
  env NUTCH_HOME=`/root/nutch_source/runtime/local`
  ```
- why not put symlinks to bin/nutch (and bin/crawl) into /usr/local/bin to make them available from everywhere?
  ```
  ln -sf $NUTCH_HOME/bin/nutch /usr/local/bin/
  ln -sf $NUTCH_HOME/bin/crawl /usr/local/bin/
 ```

lewismc: Excellent points @sebastian-nagel I'll implement them. Thanks

PR #697, opened 2021-09-21
NUTCH-2896 Protocol-okhttp: make connection pool configurable

sebastian-nagel:
- add configuration property `http.connection.pool.okhttp` to configure the number of connection pools, their size and the keep-alive time of the pooled connections
- create as many clients as pools are configured, each client holding one pool
- distribute connections by target host name over clients

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -428,6 +428,27 @@
   </description>
 </property>
 
+<property>
+  <name>http.connection.pool.okhttp</name>
+  <value></value>
+  <description>
+    (EXPERT) List of 3 integer values [M,N,T]: Instantiate
+    protocol-okhttp with M connection pools of size N and a keep-alive
+    time of T seconds.  The product M*N should approx. correspond to
+    the number of crawled hosts in one fetcher task during the
+    keep-alive time.  Because OkHttp's connection pool (v4.9.1) is not
+    optimized for fast look-up of connections, the pool size N should
+    not exceed 1000.  To allow for efficient pooling, multiple pools
+    are instantiated and connections are distributed over the pools by
+    target host name.  If the property is undefined or empty a single
+    pool is used with OkHttp's default size and keep-alive time. For
+    OkHttp 4.9.1 the default is equivalent to a configuration value
+    &quot;1,5,300&quot; - one pool with 5 idle connections and 300
+    sec. keep-alive time.  See also NUTCH-2896 and
+    https://square.github.io/okhttp/3.x/okhttp/okhttp3/ConnectionPool.html
+  </description>
+</property>
+
 <!-- FTP properties -->
 
 <property>
--- a/src/plugin/protocol-okhttp/src/java/org/apache/nutch/protocol/okhttp/OkHttp.java
+++ b/src/plugin/protocol-okhttp/src/java/org/apache/nutch/protocol/okhttp/OkHttp.java
@@ -32,6 +32,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
@@ -51,6 +52,7 @@ import org.slf4j.LoggerFactory;
 
 import okhttp3.Authenticator;
 import okhttp3.Connection;
+import okhttp3.ConnectionPool;
 import okhttp3.Headers;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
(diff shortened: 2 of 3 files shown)


sebastian-nagel:
## Tuning the connection pool configuration

Just to share the experience tuning the OkHttp connection pool(s) for a running crawl with the following characteristics:

- one crawl is split over 100 segments and each of the 100 segments is fetched in one MapReduce job. The job configuration can be adapted which allowed to try various connection pool configurations.
- every segment is fetched within 3 hours using 40 fetcher tasks, each with about 160 fetcher threads
- there is a polite delay of minimum 6 seconds between successive fetches from the same host
- every fetcher task is assigned a shuffled fetch list of about 900k URLs, about 70% of the list are successfully fetched (HTTP status code 200). HTTP redirects are added to the fetch lists on the fly.
- 90% of the fetched URLs are from hosts with at least 30 items in a fetch list. 75% with 100 or more items. So, a well-configured connection pool can save time for the bulk of the fetches.
- for 30% of the hosts only a single URL is fetched. That's why connections shouldn't kept alive for too long. But these single-URL hosts make just 1% of the URLs in a fetch list.
- the number of unique hosts per fetcher task is about 20k with 5k hosts accessed every minute

The number of accessed hosts defines the connection pool size: for the given crawl the total size should be at least 5k and there is no need for pooling 20k connections or more.

Now a couple of trials were necessary to get to a efficient configuration. For every trial the following metrics were recorded:
- CPU time (according to MapReduce job/task counters)
- profiling few fetcher map tasks using [async-profiler](https://github.com/jvm-profiling-tools/async-profiler)

Other metrics were traced but without showing significant changes. For example, the number of successfully fetched pages didn't change more than the random variation (within the range of 29.9 and 30.0 million pages per segment).

Note: All trials were run with HTTP/2 disabled, using only HTTP/1.1 - testing with HTTP/2 is planned in a later experiment.

| job CPU hours | map tasks CPU hours | pool config | description |
|---------------|---------------------|-------------|-------------|
| 169           |                 108 |     1,5,300 | 1 pool with 5 idle connections, 300 sec. keep-alive time (baseline - OkHttp's default using a single client) |
| 172           |                 111 |    1,50,300 | 1 pool with 50 idle connections, 300 sec. keep-alive time |
| 183           |      (not measured) |  1,8192,300 | 1 pool with 8192 idle connections, 300 sec. keep-alive time |
| 141           |                  78 |  48,256,300 | 48 pools each with 256 idle connections, 300 sec. keep-alive time |
| 132           |                  68 |  64,256,300 | 64 pools each with 256 idle connections, 300 sec. keep-alive time |
| 133           |                  68 | 128,256,300 | 128 pools each with 256 idle connections, 300 sec. keep-alive time |

The profiler gives a more detailed picture where the CPU time is spent. The absolute numbers (amount of samples) of the profiler runs are not comparable because of different contexts - other tasks running on the same node, profiling time, etc. But the relative numbers should be comparable and can be compared with the time spent in methods such as the MD5 calculation or Tika's MIME detection: if more relative time is spent there, the HTTP layer needs less CPU. Below three profiling flamegraphs (or the interactive flamegraphs (HTML) zipped: [fetcher.map.okhttp-pool-html.zip](https://github.com/apache/nutch/files/7224685/fetcher.map.okhttp-pool-html.zip)):

1. the baseline - a single pool with 5 idle connections, 300 sec. keep-alive time
fetcher map okhttp-pool-1-5

2. 1 pool with 8192 idle connections, 300 sec. keep-alive time
fetcher map okhttp-pool-1-8192

3. 64 pools each with 256 idle connections each, 300 sec. keep-alive time
fetcher map okhttp-pool-64-256

The profiles show that searching for a connection in the pool is inefficient if only a single connection pool is used. Cf. a [discussion on stackoverflow](https://stackoverflow.com/questions/63047533/connection-pool-okhttp) which confirms this finding.

Conclusion: The OkHttp protocol plugin is now configured to use 64 pools (128 seems to also work well) with max. 256 idle connections per pool. The savings are impressive:
- 20% less CPU usage (169h -> 132h) for the entire job including the reduce tasks which do the WARC packaging which is also CPU intensive
- 35% less CPU usage (108h -> 68h) in the map tasks alone

lewismc:
Hi @sebastian-nagel I suggest that you take this content and create a wiki page. I need to review again. You initial conclusions look very promising.

sebastian-nagel:
@lewismc: yes, I'll add first a page about HTTP protocol plugins to the wiki. It should include a feature matrix what's supported - authentication, http/2, proxies, connection pool, etc.

sebastian-nagel:
@lewismc: I've linked this issue on a new wiki page [ProtocolImplementations](https://cwiki.apache.org/confluence/display/NUTCH/ProtocolImplementations). Merging now - the connection pool is tested since almost one year now.

PR #704, opened 2021-11-18
NUTCH-2905 Mask sensitive strings in log output of index writers

sebastian-nagel:
- add utility methods (StringUtil) to mask password strings or passwords in strings
- mask passwords in log output of index writers (Elasticsearch, Solr, RabbitMQ)
- mask password in trace log of protocol-httpclient when using basic authentication


lewismc (approved):
+1 excellent coverage. Thanks for taking care of other implementations as well.

PR #707, opened 2021-11-22
NUTCH-2908 Log mapreduce job messages and counters in local mode (Log4j2)

sebastian-nagel:


lewismc (approved):
Tested right now as I am setting up logging for Enterprise Splunk. Very nice +1

PR #708, opened 2021-11-30
fix for NUTCH-2911 contributed by prakharchaube

prakharchaube:
Fix for [NUTCH-2911](https://issues.apache.org/jira/browse/NUTCH-2911)

Changes made : Added a wrapper try-finally block for FetcherRun.run() and called cleanup()

### DIFF
--- a/src/java/org/apache/nutch/fetcher/Fetcher.java
+++ b/src/java/org/apache/nutch/fetcher/Fetcher.java
@@ -191,7 +191,8 @@ public class Fetcher extends NutchTool implements Tool {
     }	  
 
     @Override
-    public void run(Context innerContext) throws IOException {
+    public void run(Context innerContext)
+        throws IOException, InterruptedException {
 
       setup(innerContext);
       try {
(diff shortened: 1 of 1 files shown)


sebastian-nagel:
Thanks @prakharchaube, good catch. The compilation fails with `Fetcher.java:440: error: unreported exception InterruptedException; must be caught or declared to be thrown` - could you update the PR to catch the exception thrown by the [cleanup](https://hadoop.apache.org/docs/r3.3.1/api/org/apache/hadoop/mapreduce/Mapper.html#cleanup-org.apache.hadoop.mapreduce.Mapper.Context-) method. Thanks!

prakharchaube:
@sebastian-nagel I have added a catch for InterruptedException, sorry for missing it. I know ideally, we should not catch InterruptedException and simply log it, but the cleanup() call will never throw it in the first place.
Let me know if there are any concerns around this. Thanks.

sebastian-nagel:
Hi @prakharchaube, after a second look (sorry about it), you're right.

> ideally, we should not catch InterruptedException

Definitely. The overridden [Mapper run](https://hadoop.apache.org/docs/r3.3.1/api/org/apache/hadoop/mapreduce/Mapper.html#run-org.apache.hadoop.mapreduce.Mapper.Context-) method declares the InterruptedException to be thrown, so the overriding method should just throw it as well. What do you think?

prakharchaube:
Yes @sebastian-nagel I do feel that would be the best way to handle it, I noticed even job.waitForCompletion() throws InterruptedException so propagation is intentional, I will make the change and update, thanks.

prakharchaube:
Done @sebastian-nagel, please review. Thanks!

sebastian-nagel (approved):
Thanks, @prakharchaube! Looks good. Waiting for the CI build and eventual further comments. If there are no objections, will commit soon.

PR #723, opened 2022-01-14
NUTCH-2935 DeduplicationJob: failure on URLs with invalid percent encoding

sebastian-nagel:
- catch IllegalArgumentException when unescaping percent-encoding in URLs
- if one URL of two compared URLs is valid, keep it as non-duplicate
- add unit tests for DeduplicationJob


lewismc (approved):
This is real nice. I must admit  I never encountered the exception before but I suspect that has to do with the nature of the URLs we encounter within the workplace. 
Thanks for demonstrating the usage within the unit test.

sebastian-nagel:
Thanks, @lewismc! For the Common Crawls it just took some time to hit a bad URL which also somehow slipped through the URL filters and normalizers during injection.

PR #724, opened 2022-01-15
NUTCH-2573 Suspend crawling if robots.txt fails to fetch with 5xx status

sebastian-nagel:
- add properties
  - `http.robots.503.defer.visits` :
    enable/disable the feature (default: enabled)
  - `http.robots.503.defer.visits.delay` :
    delay to wait before the next trial to fetch the properties
    (default: wait 5 minutes)
  - `http.robots.503.defer.visits.retries` :
    max. number of retries before giving up and dropping all URLs from the given host / queue
    (default: give up after the 3rd retry, ie. after 4 attempts)
- handle HTTP 5xx in robots.txt parser
- handle delay, retries and dropping queues in Fetcher

Stop queuing fetch items if timelimit is reached
- re-queuing items where the robots.txt request returned a 5xx
- redirects (http.redirect.max > 0) or
- outlinks (fetcher.follow.outlinks.depth > 0)

In a first version, I forgot to verify whether the Fetcher timelimit (`fetcher.timelimit.mins`) was already reached before re-queuing the fetch item. This caused very few fetcher task to end up in an infinite loop. In detail, this happened:
1. fetcher thread starts fetching an item and requests the corresponding robots.txt. Ev., the server responds slowly.
2. fetcher timelimit is reached, all fetcher queues are flushed
3. robots.txt response "arrived". Because it's a 5xx the fetch item is re-queued and the fetch is delayed for 30 min. (custom configuration).

Then steps 1 and 3 are retried until the max number of retries is reached. But this was fixed and I've also made sure that redirects or outlinks are not queued if the timelimit is reached.

### DIFF
--- a/src/java/org/apache/nutch/fetcher/FetchItemQueues.java
+++ b/src/java/org/apache/nutch/fetcher/FetchItemQueues.java
@@ -195,7 +195,11 @@ public class FetchItemQueues {
     return null;
   }
 
-  public boolean timelimitReached() {
+  /**
+   * @return true if the fetcher timelimit is defined and has been exceeded
+   *         ({@code fetcher.timelimit.mins} minutes after fetching started)
+   */
+  public boolean timelimitExceeded() {
     return timelimit != -1 && System.currentTimeMillis() >= timelimit;
   }
 
@@ -203,7 +207,7 @@ public class FetchItemQueues {
   public synchronized int checkTimelimit() {
     int count = 0;
 
-    if (timelimitReached()) {
+    if (timelimitExceeded()) {
       // emptying the queues
       count = emptyQueues();
 
@@ -283,6 +287,7 @@ public class FetchItemQueues {
     return 0;
   }
 
+  /** See {@link #checkExceptionThreshold(String, int, long)} */
   public int checkExceptionThreshold(String queueid) {
     return checkExceptionThreshold(queueid, this.maxExceptionsPerQueue, -1);
   }
--- a/src/java/org/apache/nutch/fetcher/FetchItemQueues.java
+++ b/src/java/org/apache/nutch/fetcher/FetchItemQueues.java
@@ -287,7 +287,16 @@ public class FetchItemQueues {
     return 0;
   }
 
-  /** See {@link #checkExceptionThreshold(String, int, long)} */
+  /**
+   * Increment the exception counter of a queue in case of an exception e.g.
+   * timeout; when higher than a given threshold simply empty the queue.
+   * 
+   * @see #checkExceptionThreshold(String, int, long)
+   * 
+   * @param queueid
+   *          queue identifier to locate and check
+   * @return number of purged items
+   */
   public int checkExceptionThreshold(String queueid) {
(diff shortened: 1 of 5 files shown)


src/java/org/apache/nutch/fetcher/FetchItemQueues.java
```diff
@@ -195,11 +195,15 @@ public synchronized FetchItem getFetchItem() {
     return null;
   }
 
+  public boolean timelimitReached() {
```
lewismc, line 198: Maybe provide basic Javadoc?

src/java/org/apache/nutch/fetcher/FetchItemQueues.java
```diff
@@ -263,6 +283,10 @@ public synchronized int checkExceptionThreshold(String queueid) {
     return 0;
   }
 
+  public int checkExceptionThreshold(String queueid) {
```
lewismc, line 286: Same here. Basic Javadoc?

src/java/org/apache/nutch/fetcher/FetcherThread.java
```diff
@@ -312,6 +322,24 @@ public void run() {
...
+                fetchQueues.finishFetchItem(fit, true);
+              } else {
+                fetchQueues.addFetchItem(fit);
+              }
+              // but check whether it's time to cancel the queue
+              int killedURLs = fetchQueues.checkExceptionThreshold(
+                  fit.getQueueID(), this.robotsDeferVisitsRetries + 1,
+                  this.robotsDeferVisitsDelay);
+              if (killedURLs != 0) {
+                context.getCounter("FetcherStatus",
```
lewismc, line 338: Can you please augment https://cwiki.apache.org/confluence/display/NUTCH/Metrics

src/java/org/apache/nutch/fetcher/FetcherThread.java
```diff
@@ -600,6 +628,12 @@ private FetchItem queueRedirect(Text redirUrl, FetchItem fit)
       LOG.debug(" - ignoring redirect from {} to {} as duplicate", fit.url,
           redirUrl);
       return null;
+    } else if (fetchQueues.timelimitReached()) {
+      redirecting = false;
+      context.getCounter("FetcherStatus", "hitByTimeLimit").increment(1);
```
lewismc, line 633: Same with this one https://cwiki.apache.org/confluence/display/NUTCH/Metrics

lewismc (changes requested):
I like this patch. It's actually something which will help us loop back in with website administrators to notify them of service issues. Thanks for including the metrics this is really useful. Some minor suggestions from me @sebastian-nagel .

sebastian-nagel:
Hi @lewismc, done: updated metrics wiki page (hitByTimeLimit is already documented), added Javadocs and renamed the counter to follow the naming convention of the other robots_* counters. Also renamed the method ("timelimitReached" -> "timelimitExceeded").

lewismc:
Looks like it failed on Javadoc generation @sebastian-nagel

lewismc (approved):
Once Javadoc fixed +1 from me.

PR #725, opened 2022-01-16
NUTCH-2938 Use Any23's RepositoryWriter to write structured data to Rdf4j repository

lewismc:
PR addresses https://issues.apache.org/jira/browse/NUTCH-2938
We could improve the performance of this plugin if we could reuse the repository connection however I am not entirely sure how to do that right now because this is done down in the Any23 layer.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -976,15 +976,6 @@
 </property>
 
 <property>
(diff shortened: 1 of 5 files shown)


lewismc:
Closing this off to clean up the PR list. We are planning on retiring Any23 soon.

PR #726, opened 2022-01-16
NUTCH-2936 Early registration of URL stream handlers provided by plugins may fail Hadoop jobs running in distributed mode

lewismc:
I ended up producing this PR as a result of investigating NUTCH-2936. This PR does not fix NUTCH-2936.
The problem is that the [trustAllSslSocketFactory](https://github.com/apache/nutch/blob/master/src/plugin/protocol-okhttp/src/java/org/apache/nutch/protocol/okhttp/OkHttp.java#L129) variable is `null` when passed into `okhttp3.OkHttpClient.Builder.sslSocketFactory`.
I have not yet tried to verify if running protocol-okhttp is distributed mode. I can try that early next week. It might be worthwhile us reverting NUTCH-2429.

### DIFF
--- a/src/java/org/apache/nutch/parse/ParserChecker.java
+++ b/src/java/org/apache/nutch/parse/ParserChecker.java
@@ -114,15 +114,15 @@ public class ParserChecker extends AbstractChecker {
     int numConsumed;
     for (int i = 0; i < args.length; i++) {
       if (args[i].equals("-normalize")) {
-        normalizers = new URLNormalizers(getConf(), URLNormalizers.SCOPE_DEFAULT);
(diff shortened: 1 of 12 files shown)


sebastian-nagel:
+1 Afaics, this PR addresses only code style, code conventions, Javadoc, etc. but does not change anything functionally. Maybe this should be reflected in the commit messages as well.

sebastian-nagel (approved):

lewismc:
Thanks @sebastian-nagel 
Again, this DOES NOT fix the underlying issue. I will investigate that again when I get the free cycles.

PR #730, opened 2022-05-05
NUTCH-2948 Upgrade dependencies to Any23 2.7 and Tika 2.3.0

sebastian-nagel:


lewismc (approved):
+1 tested on lots of HTML :)
I see that Tika 2.4.0 is release and I literally just performed the upgrade in Any23... we can perform the next upgrade soon enough. Thanks

PR #731, opened 2022-05-19
NUTCH-2950 UpdateHostDb: performance improvements

sebastian-nagel:
(see NUTCH-2950 and commit messages)


sebastian-nagel:
Updated PR:
- fix the Javadoc errors
- fail "javadoc" target on all kinds of errors/warnings independent from the system configuration

PR #734, opened 2022-06-15
NUTCH-2952 Upgrade core dependencies

sebastian-nagel:
Upgrade of core dependencies
- Hadoop 3.1.3 -> 3.3.3
- log4j 2.17.0 -> 2.17.2
- and some more

Note: I've observed that some unit tests are failing with same/similar errors than observed in NUTCH-2936. Eventually, this PR needs to be rebased on top of #733.

### DIFF
--- a/ivy/ivy.xml
+++ b/ivy/ivy.xml
@@ -36,10 +36,10 @@
 	</publications>
 
 	<dependencies>
-		<dependency org="org.apache.logging.log4j" name="log4j-api" rev="2.17.0" conf="*->master" />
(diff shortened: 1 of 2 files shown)


sebastian-nagel:
Update: the failing unit test (TestCrawlDbDeduplication) on my development system stem from a modified nutch-site.xml requesting protocol-okhttp - obviously, it's the combination of using protocol-okhttp and a more recent Hadoop version which triggers the appearance of NUTCH-2936: the issue (NoSuchAlgorithmException) is then reproducible also in local mode.

PR #736, opened 2022-06-23
NUTCH-2930 Protocol-okhttp: implement IP filter

sebastian-nagel:
Implement filter by IP address for protocol-okhttp, configurable via include/exclude rules as list of IP address, CIDR notation
  or predefined IP ranges (localhost, loopback, sitelocal).


sebastian-nagel:
(rebased to recent master and resolved merge conflicts)

PR #754, opened 2023-01-26
NUTCH-2982 Generator: parameter for URL normalization not passed forward

sebastian-nagel:


sebastian-nagel:
(close in favor of #755 - same solution anyway)

PR #777, opened 2023-09-19
NUTCH-3002 Protocol-okhttp HttpResponse: HTTP header metadata lookup should be case-insensitive

sebastian-nagel:
- implement class CaseInsensitiveMetadata providing case-insensitive metadata look-ups (but no spell-checking)
- use CaseInsensitiveMetadata to hold HTTP header metadata in in the class OkHttpResponse of protocol-okhttp
- add unit tests to prove the fix (and also case-insensitive look-ups and spell-checking in protocol-http)

The unit tests also include a deactivated benchmark: storing HTTP headers in CaseInsensitiveMetadata is approx. 10% faster than in SpellCheckedMetadata.


jnioche:

PR #779, opened 2023-09-26
NUTCH-2990 HttpRobotRulesParser to follow 5 redirects as specified by RFC 9309

sebastian-nagel:
- follow multiple redirects when fetching robots.txt
- number of followed redirects is configurable by the property `http.robots.redirect.max` (default: 5)
- improvements in RobotRulesParser's robots.txt test utility
  - bug fix: the passed agent names need to be transferred to the property http.robots.agents earlier, before the protocol plugins are configured
  - more verbose debug logging

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -164,6 +164,16 @@
 </property>
 
 <property>
+  <name>http.robots.redirect.max</name>
+  <value>5</value>
+  <description>Maximum number of redirects followed when fetching
+  a robots.txt file. RFC 9309 specifies that &quot;crawlers SHOULD
+  follow at least five consecutive redirects, even across authorities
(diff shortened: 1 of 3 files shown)


lewismc:
Very nice @sebastian-nagel 
Do you have an example on hand of a robots.txt which can be fetched with >1 redirects?

sebastian-nagel:
>  an example on hand of a robots.txt which can be fetched with >1 redirects?

http://wikipedia.org/robots.txt

Note: works with protocol-http, for protocol-okhttp need also to apply the fix for NUTCH-3002.

Maybe as an additional note: this PR removes the secondary lookup for a lower-cased "location" header. Case-insensitive lookup of protocol metadata should be implemented on protocol level.

jnioche:

PR #784, opened 2023-09-30
NUTCH-2897 Do not supress deprecated API warnings

sebastian-nagel:
- deprecate constructor of NutchJob
- remove deprocated call to Object.finalize() from Plugin.finalize()


lewismc (approved):
+1

PR #786, opened 2023-10-01
NUTCH-3011 HttpRobotRulesParser: handle HTTP 429 Too Many Requests same as server errors (HTTP 5xx)

sebastian-nagel:


jnioche:

PR #788, opened 2023-10-21
NUTCH-3013 Employ commons-lang3's StopWatch to simplify timing logic

lewismc:
Addresses https://issues.apache.org/jira/browse/NUTCH-3013

If this gets merged I'll squash all of those commits. 

Note, this does change the logging slightly. The whole `LOG.info("CrawlDb update: starting at {}", sdf.format(start));` is redundant because we already capture the datetime stamp in the log output anyway.
I simplified this to `LOG.info("CrawlDb update: starting");`.

Passes local unit tests.


lewismc:
If this gets merged I'll squash all of those commits. 

Note, this does change the logging slightly. The whole `LOG.info("CrawlDb update: starting at {}", sdf.format(start));` is redundant because we already capture the datetime stamp in the log output anyway.
I simplified this to `LOG.info("CrawlDb update: starting");`.

lewismc:
Passes local unit tests.

sebastian-nagel (approved):
+1 looks good, tested a couple of tools

PR #792, opened 2023-10-30
Allow fast-urlfilter to load from HDFS/S3 and support gzipped input [NUTCH-3017]

jnioche:
See description in https://issues.apache.org/jira/browse/NUTCH-3017


jnioche:
Obivously, pulled more changes than I meant to

PR #796, opened 2023-11-06
[NUTCH-3025] urlfilter-fast to filter based on the length of the URL

jnioche:
Checks the length of the query and path elements of a URL. 
see https://issues.apache.org/jira/browse/NUTCH-3025

### DIFF
--- a/src/plugin/urlfilter-fast/src/java/org/apache/nutch/urlfilter/fast/FastURLFilter.java
+++ b/src/plugin/urlfilter-fast/src/java/org/apache/nutch/urlfilter/fast/FastURLFilter.java
@@ -97,6 +97,7 @@ public class FastURLFilter implements URLFilter {
 
   private Configuration conf;
   public static final String URLFILTER_FAST_FILE = "urlfilter.fast.file";
+  public static final String URLFILTER_FAST_MAX_LENGTH = "urlfilter.fast.url.max.length";
   public static final String URLFILTER_FAST_PATH_MAX_LENGTH = "urlfilter.fast.url.path.max.length";
   public static final String URLFILTER_FAST_QUERY_MAX_LENGTH = "urlfilter.fast.url.query.max.length";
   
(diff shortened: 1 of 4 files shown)


src/plugin/urlfilter-fast/src/java/org/apache/nutch/urlfilter/fast/FastURLFilter.java
```diff
@@ -97,9 +97,17 @@ public class FastURLFilter implements URLFilter {
 
   private Configuration conf;
   public static final String URLFILTER_FAST_FILE = "urlfilter.fast.file";
+  public static final String URLFILTER_FAST_PATH_MAX_LENGTH = "urlfilter.fast.url.path.max.length";
+  public static final String URLFILTER_FAST_QUERY_MAX_LENGTH = "urlfilter.fast.url.query.max.length";
+
```
sebastian-nagel, line 102: What about adding a third limit for path and query combined ([URL.getFile()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URL.html#getFile())?
- if somebody defined two generous but reasonable limits (for example, 2048) for both path and query, the resulting URL may still get quite long and cause troubles 
- also the HTTP GET request includes both path and query
- for many use cases it should sufficient to just set this limit

sebastian-nagel (changes requested):
Thanks, @jnioche. Looks great!

One point more: all properties should be listed and documented in conf/nutch-default.xml

src/plugin/urlfilter-fast/src/java/org/apache/nutch/urlfilter/fast/FastURLFilter.java
jnioche: I might keep things simple and just add a size limit on the whole URL regardless of its parts, similar to [what is done in StormCrawler.](https://github.com/DigitalPebble/storm-crawler/blob/ef31e509139cccb2919c345ef343c4fcfb2f1ec5/core/src/main/java/com/digitalpebble/stormcrawler/filtering/basic/BasicURLFilter.java#L30C17-L30C26)

jnioche:
Writing a test for this thing is an absolute pain. The way the filters are used for real is that their method setConf is called and the rules are loaded using _getConfResourceAsReader_, i.e. they are expected to be in the jar.
The tests do not rely on that mechanism and instead instantiate the filter with the reader for its rules. This means that the conf is not used at all and therefore we can't use that to load the value for the length based filters. I will add another constructor with the reader + conf so that we can test based on the length.

sebastian-nagel (approved):
+1 Looks good. Tested successfully using the filterchecker.

Note: there's a merge conflict after merging NUTCH-3017 into master. Could you resolve it?

jnioche:
@sebastian-nagel merged the changes from master and made a few improvements

sebastian-nagel:
Thanks, @jnioche!

PR #557, opened 2020-12-08
[NUTCH-2834] Add possibility to setup deduplication group mode in crawl script

derhecht:

PR #800, opened 2023-12-14
[NUTCH-2834] Update crawl documentation / Fix #557

derhecht:
Show --dedup-group instead of -dedup-group which have lead to misunderstanding output


sebastian-nagel:
Thanks, @derhecht! Good catch!

PR #801, opened 2023-12-14
Update Dockerfile / JAVA_HOME

derhecht:
Alpine is using ash shell by default which results in an not set JAVA_HOME environment variable

Sry, there is no issue reported atm on issues.apache.org - never the less, it is one I'm facing to


derhecht:
see #805

PR #802, opened 2024-01-18
fix for NUTCH-3027 contributed by skehrli

skehrli:
Fix for NUTCH-3027: trivial resource leak in .... /util/DomainSuffixes.java


sebastian-nagel:
Patch applied to master in d95e1a7, see comments on Jira in NUTCH-3027. Thanks again @skehrli !

PR #803, opened 2024-03-11
NUTCH-3033 Upgrade Ivy to v2.5.2

lewismc:
PR for https://issues.apache.org/jira/browse/NUTCH-3033
I was having trouble locally resolving the Ivy version to 2.5.2… I can’t yet figure out why 2.5.1 was being used. I’ll check out the CI log and see if the newer version is used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,7 @@ build/
 runtime/
 logs/
 /bin/
-ivy/ivy-2.3.0.jar
-ivy/ivy-2.4.0.jar
-ivy/ivy-2.5.0-rc1.jar
-ivy/ivy-2.5.0.jar
+ivy/ivy-2.*
 ivy/spotbugs-*/
 naivebayes-model
 .naivebayes-model.crc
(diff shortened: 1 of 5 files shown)


lewismc:
OK so it looks like the [newer Ivy version is being used just fine](https://github.com/apache/nutch/actions/runs/8239165168/job/22531780061?pr=803#step:4:78). The build did however fail with the following
```
Caused by: java.lang.RuntimeException: Multiple artifacts of the module io.netty#netty-transport-native-kqueue;4.1.84.Final are retrieved to the same file! Update the retrieve pattern to fix this error.
```
… investigating.

lewismc:
Hmmm, I upgraded to 2.5.1 and the CI runs just fine. Looks like there is some regression/additional configuration required with 2.5.2. I’m asking the question over on ivy-user@ mailing list.

lewismc:
After lots of trial and error I think I cracked this one. Ultimately there were several places where the optional `(-[classifier])` element has to be added to the `ivy:retrieve pattern`. 
This wasn’t particularly intuitive as the ivy documentation is [somewhat lacking in this regard](https://ant.apache.org/ivy/history/2.5.2/resolver/filesystem.html#_child_elements) however @bodewig [pointed me in the right direction on the ivy-user@ mailing list](https://lists.apache.org/thread/fdd9r5gkdk5215hc9swcxhjwyvnzoz0w). Thank you for that @boedwig.

This PR is ready for review.

lewismc:
Thanks @sebastian-nagel 👍

PR #805, opened 2024-03-12
Update Dockerfile / JAVA_HOME - 2nd try

derhecht:
Alpine is using ash shell by default which results in an not set JAVA_HOME environment variable

Sry, there is no issue reported atm on issues.apache.org - never the less, it is one I'm facing to

see #801


lewismc:
Thanks @derhecht 👍

PR #815, opened 2024-04-25
NUTCH-3044 Generator: NPE when extracting the host part of a URL fails

sebastian-nagel:


lewismc (approved):
+1 LGTM

I was thinking we could provide a `TestGenerator#testNullHostInReducer` test case which would bootstrap the `SelectorReducer.reduce` with some "well formed" and "not well-formed" URLs in order to validate that no NPE is thrown. The test case should fail before your patch is applied and pass afterwards.

sebastian-nagel:
> we could provide a TestGenerator#testNullHostInReducer test case

Good idea! Done, see 4729786.

sebastian-nagel:
... also fixed the Javadoc error.

lewismc:
Excellent @sebastian-nagel +1

PR #817, opened 2024-04-30
NUTCH-3054 Address deprecation of Node16 for all GitHub Actions

lewismc:
Addresses https://issues.apache.org/jira/browse/NUTCH-3054


sebastian-nagel (approved):
+1

PR #821, opened 2024-07-09
NUTCH-3061 URL filters to log name of the rules file

sebastian-nagel:
Let urlfilter-regex, urlfilter-automaton and urlfilter-fast log the name of the rules file. Also log the number of rules read from the file.

Cleanup (Javadoc) comments in the modified classes.


lewismc:
+1 @sebastian-nagel applied cleanly against `master`. I think this logging is certainly useful.

PR #824, opened 2024-09-06
NUTCH-3066 Protocol plugin unit tests fail randomly

sebastian-nagel:
- use random free ephemeral port in AbstractHttpProtocolPluginTest
- store used port in field "serverPort", inherited by plugin tests
- use serverPort to sleep longer until server is up and set its port number


lewismc:
+1 @sebastian-nagel applied and tested against `master`.

PR #827, opened 2024-10-04
NUTCH-3067 Improve performance of FetchItemQueues if error state is preserved

sebastian-nagel:
Address NUTCH-3067:

1. do not keep every stateful queue: drop queues which have a low exception count after a configurable amount of time. If a second URL from the same host/domain/IP is fetched after a considerably long time span (eg. 30 minutes), the effect on performance and politeness should be negligible.

   This is configured by `fetcher.exceptions.per.queue.clear.after` (default 30 minutes): if this time has elapsed after the next fetch time in the queue, that is in addition to any delay defined by the exponential backoff, empty queues are dropped.

2. reviewed and improved the handling of the exponential backoff in [FetchQueues.checkExceptionThreshold](https://nutch.apache.org/documentation/javadoc/api/org/apache/nutch/fetcher/FetchItemQueues.html#checkExceptionThreshold(java.lang.String,int,long)): if the delayed next fetch would happen after the fetcher timelimit (if configured), the queue is purged and blocked because no fetch item will be fetched from the queue anyway.

3. reduce the memory footprint of a single FetchQueue - important if there are many of them.

In addition, logging and documentation has been improved. While testing this PR on a production crawl, a NUTCH-3072 has been detected.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -1198,7 +1198,11 @@
   <name>fetcher.throughput.threshold.retries</name>
   <value>5</value>
   <description>The number of times the fetcher.throughput.threshold.pages is allowed to be exceeded.
-  This settings prevents accidental slow downs from immediately killing the fetcher thread.
+  This settings prevents accidental slow downs from immediately shutting down the fetcher threads.
+  The throughput is checked approx. every second. Note: the number of retries should be lower
+  than the timeout defined by mapreduce.task.timeout and fetcher.threads.timeout.divisor (see there).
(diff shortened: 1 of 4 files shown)


sebastian-nagel:
This PR is successfully tested in production: Using the default of 30 minutes for `fetcher.exceptions.per.queue.clear.after` the number of FetchQueues hold stabilizes after half an hour because the queues with a single error and no new additions of URLs to the queue are removed after this time span.

sebastian-nagel:
(rebased on recent master)

PR #829, opened 2024-10-15
NUTCH-3078 Unlock database when Injector finishes - regardless of result

HiranChaudhuri:


src/java/org/apache/nutch/crawl/Injector.java
```diff
@@ -402,40 +402,38 @@ public void inject(Path crawlDb, Path urlDir, boolean overwrite,
 
     // lock an existing crawldb to prevent multiple simultaneous updates
     Path lock = CrawlDb.lock(conf, crawlDb, false);
```
sebastian-nagel, line 404: See the discussion in NUTH-3078: moving the creating of the lockfile down just before the job is run would avoid that a failure during job configuration leads to a lost lockfile.

sebastian-nagel (changes requested):

src/java/org/apache/nutch/crawl/Injector.java
sebastian-nagel: Great!

sebastian-nagel (approved):
Thanks,  @HiranChaudhuri!

I've also verified that this solution fixes the issue with the lockfile if the seed URL directory does not exists.

lewismc (approved):
+1 nice PR @HiranChaudhuri

PR #830, opened 2024-10-19
This improves the way Nutch is erroring out, at least for local mode.

HiranChaudhuri:
Nutch will print more information to the logfile - which in local mode ends up on the terminal.
Not sure what will happen in distributed mode though.

### DIFF
--- a/src/java/org/apache/nutch/crawl/Injector.java
+++ b/src/java/org/apache/nutch/crawl/Injector.java
@@ -130,23 +130,28 @@ public class Injector extends NutchTool implements Tool {
 
     @Override
     public void setup(Context context) {
-      Configuration conf = context.getConfiguration();
-      boolean normalize = conf.getBoolean(CrawlDbFilter.URL_NORMALIZING, true);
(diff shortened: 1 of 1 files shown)


src/java/org/apache/nutch/crawl/Injector.java
```diff
@@ -130,23 +130,28 @@ public static class InjectMapper
...
+        interval = conf.getInt("db.fetch.interval.default", 2592000);
+        if (filter) {
+          filters = new URLFilters(conf);
+        }
+        scfilters = new ScoringFilters(conf);
+        scoreInjected = conf.getFloat("db.score.injected", 1.0f);
+        curTime = conf.getLong("injector.current.time",
+            System.currentTimeMillis());
+        url404Purging = conf.getBoolean(CrawlDb.CRAWLDB_PURGE_404, false);
+      } catch (Exception e) {
```
lewismc, line 151: Exception is very vague. I would suggest we figure out what types of Exception's are thrown then catch and raise them instead. What do you think @HiranChaudhuri  ?

PR #831, opened 2024-10-20
NUTCH-3075 tld plugin makes injector crash

sebastian-nagel:
NUTCH-1942 Remove TopLevelDomain

- unregister TLDScoringFilter from the tld plugin.xml


lewismc (approved):
+1

PR #836, opened 2024-10-27
NUTCH-2771 Tests in nightly builds: skip long runners

sebastian-nagel:
- introduce target "test-full" which executes all unit tests, even slow ones
- slow unit tests are only run if the system property "test.include.slow" is true (it is set to true by the target "test-full")

- a solution based on JUnit tags isn't possible since we use Ant and are still on JUnit 4


lewismc (approved):
+1 I like it. As you mentioned, we can simplify this through JUnit5 annotations further down the line. Nice PR @sebastian-nagel

PR #837, opened 2024-10-27
NUTCH-3079 Dumping a segment fails unless it has been fetched and parsed

sebastian-nagel:
SegmentReaders dump and get tools do now check whether a segment subdirectory exists before adding it as input. A warning is shown if the subdirectory does not exist but is not excluded via one of the general options (-nogenerate, etc.)


lewismc (approved):
+1 
Maybe we could add Javadoc for the new methods.

PR #841, opened 2024-11-27
NUTCH-3094 Github tests to run if build configuration changes

sebastian-nagel:
This PR improves the logic of conditional steps in the Github workflow:
- all tests are executed if any of build configuration, core or plugins code changed
- core tests if there are only changes in core
- plugin tests if there are only changes in the plugins source folder

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -82,11 +82,18 @@ jobs:
               - 'src/testresources/**'
             plugins:
               - 'src/plugin/**'
+            buildconf:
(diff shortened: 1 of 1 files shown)


sebastian-nagel:
Ok, this didn't work as expected: all tests were run although there were no changes in build configuration or source code. Need to improve the PR...

lewismc (approved):
+1 
We could possibly extend this further to run dependency-related targets if any ivy configuration changes but that is outside the scope.
Nice patch

PR #848, opened 2025-02-06
[NUTCH-3103] Fixed custom max intervals for AdaptiveFetchSchedule

martin-djukanovic:
1) The loop in setHostSpecificIntervals is cleaned up and if max interval in the config is set to default, it is treated correctly.
2) The functions getMinInterval and getMaxInterval are respectively renamed to getCustomMinInterval and getCustomMaxInterval and now return null if no custom interval has been set for the given URL's hostname. If one of them returns null after it is called, then the corresponding default value will be used to bound the calculated interval.
3) The custom interval values in the config are now allowed to equal the default values. For example, if the default min interval is 7200 then in the config file "0", "default" and "7200" are all valid values for the custom min interval, and they all have the same result.
4) The config file template is changed to account for these changes.


sebastian-nagel:
Merged. Thanks, @martin-djukanovic!

PR #856, opened 2025-07-11
Nutch 3115 Extend access to all filter plugin constructor call args to user's POJO in Arbitrary Indexer

CatChullain:
This is the update to the Arbitrary Indexing Filter plugin that lets it optionally pass all its own constructor args (NutchDocument, Parse, Text (from Hadoop), CrawlDatum, and Inlinks) to the user's POJO constructor for their custom processing. There's a small typo correction in a logging statement, too, that output a class name when it claimed to output a method name, included here at no extra cost! :)

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -2337,19 +2337,19 @@ CAUTION: Set the parser.timeout to -1 or a bigger value than 30, when using this
 
 <property>
   <name>index.arbitrary.overwrite.0</name>
+  <value></value>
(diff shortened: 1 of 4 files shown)


CatChullain:
Still banging on this today. If we still have time, I'd like to do a bit more work so the new flag for all fields access can apply to each field/class individually. That'll be easier to update an existing nutch-site to start using it without interfering with previous index-arbitrary defs.

sebastian-nagel (approved):
Thanks, @CatChullain! I'm going to merge the PR...

PR #859, opened 2025-08-19
[NUTCH-3122] Make SpellCheckedMetadata case-insensitive for all Metadata names

TamimEhsan:
PR for https://issues.apache.org/jira/browse/NUTCH-3122

The `SpellCheckedMetadata` class adds spellchecking capabilities to some fixed HTTP headers. However, other metadatas remain case sensitive. By extending `CaseInsensitiveMetadata` instead of `Metadata` provides the case insensitivity features.

### DIFF
--- a/src/java/org/apache/nutch/metadata/SpellCheckedMetadata.java
+++ b/src/java/org/apache/nutch/metadata/SpellCheckedMetadata.java
@@ -26,7 +26,7 @@ import org.apache.commons.lang3.StringUtils;
 /**
  * A decorator to Metadata that adds spellchecking capabilities to property
  * names. Currently used spelling vocabulary contains just the HTTP headers from
- * {@link HttpHeaders} class.
+ * {@link HttpHeaders} class. Other names are case insensitive.
  * 
  */
 public class SpellCheckedMetadata extends CaseInsensitiveMetadata {
--- a/src/test/org/apache/nutch/metadata/TestSpellCheckedMetadata.java
+++ b/src/test/org/apache/nutch/metadata/TestSpellCheckedMetadata.java
@@ -21,6 +21,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.Properties;
 
 import org.junit.Assert;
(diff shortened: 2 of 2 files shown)


lewismc:
Thanks for the PR @TamimEhsan 😄 
This patch will be backwards compatible with existing DB's but it would be great if we could possibly test for that as well.
Out of curiosity can you please provide a use case which motivated you to develop this patch?
Thank you

sebastian-nagel:
Thanks, @TamimEhsan! 

To add more details to @lewismc's comment: the "Metadata" class implements "Writable" because it needs to be serialized when data is stored in the CrawlDb and in segments, or when Nutch is run on a Hadoop cluster and data is exchanged between distributed tasks. Ideally, serialization is backward-compatible, that is Metadata written before this change is readable afterwards.

> make any comments on the Jira ticket

If you want, you can request a Apache Jira account [here](https://selfserve.apache.org/jira-account.html).

TamimEhsan:
is this test enough to check the backward compatibility? I tried to recreate the old behaviour by reassigning HashMap to metadata. Thus simulating a writing by old instance and reading from new. 

```java
  @Test
  public void testWritableCaseInsensitive() {
    SpellCheckedMetadata result = null;
    SpellCheckedMetadata meta = new SpellCheckedMetadata() {
      {
          this.metadata = new HashMap<>();
      }
    };

    result = writeRead(meta);
    Assert.assertEquals(0, result.size());

    meta.add("name-One", "value-1.1");
    result = writeRead(meta);
    Assert.assertEquals(1, result.size());
    Assert.assertEquals(1, result.getValues("naMe-one").length);
    Assert.assertEquals("value-1.1", result.get("nAme-oNe"));

    meta.add("Contenttype", "text/html");
    meta.add("name-Two", "value-2.1");
    meta.add("namE-two", "value-2.2");
    result = writeRead(meta);
    Assert.assertEquals(3, result.size());
    Assert.assertEquals(1, result.getValues("name-onE").length);
    Assert.assertEquals("value-1.1", result.getValues("namE-one")[0]);
    Assert.assertEquals(2, result.getValues("name-two").length);
    Assert.assertEquals("value-2.1", result.getValues("nAme-tWo")[0]);
    Assert.assertEquals("value-2.2", result.getValues("namE-Two")[1]);
    Assert.assertEquals("text/html", result.get(Metadata.CONTENT_TYPE));
  }
  ```

lewismc:
@TamimEhsan thanks for providing the additional context. You missed no steps in the contribution. The contribution is excellent thank you.
Because we own the `SpellCheckedMetadata` Class I think it is OK to mock it.
Thanks for proposing the additional test. Some documentation to accompany the test would be great.

lewismc (approved):
+1 thanks @TamimEhsan

sebastian-nagel (approved):
Thanks, @TamimEhsan! Looks good to me. Going to merge the PR.

PR #791, opened 2023-10-24
NUTCH-2887 Migrate to JUnit 5 Jupiter

lewismc:
WIP to address https://issues.apache.org/jira/browse/NUTCH-2887

PR #861, opened 2025-09-04
NUTCH-2887 Migrate to JUnit 5 Jupiter

lewismc:
# Whats in this PR?

Phase 1 of [NUTCH-2887](https://issues.apache.org/jira/browse/NUTCH-2887) which
1. introduces the JUnit 5 (Jupiter) dependencies including the `junit-vintage-engine` which allows us to run JUnit 4 tests as we gradually migrate `core` and `plugins` to JUnit 5.
1. replaces the (old) `<junit>` Ant task with `<junitlauncher` for the `test-core` Ant target.

# Next steps

## Phase 2 - core

1. Replace all JUnit 4 API usage in `core` tests with JUnit 5 implementations.
2. Investigate and use [annotations](https://docs.junit.org/current/user-guide/#writing-tests-annotations) in `core` tests which will reduce duplicate boilerplate and improve test filtering.

## Phase 3 - core

Same as in core above but replace `core` with `plugins`.

# Questions

1. Should we decompose this issue into three parts? In my opinion #791 ended up a mess and I would like to avoid that happening again. My proposal is therefore to merge this initial PR first. We then then merge phase 2 and finally phase 3. Any thoughts on this approach?
2. I'm going to also compile a list of out of scope nice to have items I uncover whilst working on this task.

Thanks for any comments. 👍

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -475,55 +475,56 @@
     </antcall>
   </target>
 
-  <target name="test-core" depends="compile-core-test, job" description="--> run core JUnit tests only">
-
+  <target name="test-core" depends="compile-core-test, job" description="--> run core JUnit tests">
     <delete dir="${test.build.data}"/>
     <mkdir dir="${test.build.data}"/>
-    <!--
-     copy resources needed in junit tests
-    -->
     <copy todir="${test.build.data}">
       <fileset dir="src/testresources" includes="**/*"/>
     </copy>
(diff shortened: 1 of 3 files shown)


lewismc:
Phase 1 is ready to test locally. Again, this PR is limited to adjusting the JUnit dependencies and associated Ant configuration in order to have the JUnit 5 engine run the existing JUnit 4 tests in `core` and `plugins`. It represents a small start. Thanks.

PR #904, opened 2026-02-26
NUTCH-3161 Address Sonarcloud High and Medium Security Hotspots

lewismc:
PR to address [NUTCH-3161](https://issues.apache.org/jira/browse/NUTCH-3161). This patch addresses following Security Hotspota
 
### High
(false positives): Exclude plugin resource directories from analysis in `sonar-project.properties`. No Java code lives in `conf`, `data`, or `sample` under `src/plugin/**`, so these paths are excluded from scanning., and 

### Medium
* ParseOutputFormat (`src/java/.../ParseOutputFormat.java`): Replace regex `" *, *"` for `db.parsemeta.to.crawldb` with comma-split + trim in `getParseMetaToCrawlDBKeys()`. Add `TestParseOutputFormat` with tests for empty, single, comma-separated, trim, and empty-segment handling.
* HtmlParser (parse-html plugin): Remove `metaPattern`, `charsetPattern`, and `charsetPatternHTML5`. Use linear string parsing in `extractCharsetFromMeta()` / `extractCharsetValue()` for HTML4 and HTML5 meta charset detection. Add `testExtractCharsetFromMeta` in existing `TestHtmlParser`.
* JSParseFilter (parse-js plugin): Remove `STRING_PATTERN` and `URI_PATTERN`. Use `extractQuotedStrings()` and `looksLikeUri()` (linear scan / simple checks). Add `testExtractQuotedStrings` and `testLooksLikeUri` in existing `TestJSParseFilter`.
* UrlValidator (urlfilter-validator plugin): Remove `URL_PATTERN` and `AUTHORITY_PATTERN`. Use `java.net.URI` for URL structure and `parseAuthority()` for host/port. Remove unused `isBlankOrNull`. Add `testParseAuthority` in existing`TestUrlValidator`.

Thanks for any review.

### DIFF
--- a/src/java/org/apache/nutch/parse/ParseOutputFormat.java
+++ b/src/java/org/apache/nutch/parse/ParseOutputFormat.java
@@ -57,6 +57,7 @@ import java.lang.invoke.MethodHandles;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map.Entry;
 
@@ -73,7 +74,23 @@ public class ParseOutputFormat extends OutputFormat<Text, Parse> {
     NUMBER_FORMAT.setMinimumIntegerDigits(5);
     NUMBER_FORMAT.setGroupingUsed(false);
   }
-  
+
+  /**
+   * Parses the comma-separated db.parsemeta.to.crawldb config value.
+   * Uses comma-split and trim to avoid ReDoS from regex backtracking.
+   * @param value config value (may be null or empty)
+   * @return array of trimmed, non-empty metadata keys (never null)
+   */
+  static String[] getParseMetaToCrawlDBKeys(String value) {
+    if (value == null || value.isEmpty()) {
+      return new String[0];
+    }
+    return Arrays.stream(value.split(",", -1))
+        .map(String::trim)
+        .filter(s -> !s.isEmpty())
+        .toArray(String[]::new);
+  }
+
   private static class SimpleEntry implements Entry<Text, CrawlDatum> {
     private Text key;
     private CrawlDatum value;
@@ -177,8 +194,8 @@ public class ParseOutputFormat extends OutputFormat<Text, Parse> {
     Path data = new Path(new Path(out, ParseData.DIR_NAME), name);
     Path crawl = new Path(new Path(out, CrawlDatum.PARSE_DIR_NAME), name);
 
-    final String[] parseMDtoCrawlDB = conf.get("db.parsemeta.to.crawldb", "")
-        .split(" *, *");
+    final String[] parseMDtoCrawlDB = getParseMetaToCrawlDBKeys(
+        conf.get("db.parsemeta.to.crawldb", ""));
 
     // textOut Options
     final MapFile.Writer textOut;
--- a/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
+++ b/src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
@@ -24,8 +24,6 @@ import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.net.URL;
 import java.net.MalformedURLException;
 import java.nio.charset.StandardCharsets;
@@ -64,15 +62,10 @@ public class HtmlParser implements Parser {
   // NUTCH-2042 (cf. TIKA-357): increased to 8 kB
   private static final int CHUNK_SIZE = 8192;
 
-  // NUTCH-1006 Meta equiv with single quotes not accepted
-  private static Pattern metaPattern = Pattern.compile(
-      "<meta\\s+([^>]*http-equiv=(\"|')?content-type(\"|')?[^>]*)>",
-      Pattern.CASE_INSENSITIVE);
-  private static Pattern charsetPattern = Pattern.compile(
-      "charset=\\s*([a-z][_\\-0-9a-z]*)", Pattern.CASE_INSENSITIVE);
-  private static Pattern charsetPatternHTML5 = Pattern.compile(
-      "<meta\\s+charset\\s*=\\s*[\"']?([a-z][_\\-0-9a-z]*)[^>]*>",
-      Pattern.CASE_INSENSITIVE);
+  private static final String META_TAG_START = "<meta";
+  private static final String CHARSET_EQ = "charset=";
+  private static final String HTTP_EQUIV = "http-equiv";
+  private static final String CONTENT_TYPE = "content-type";
 
   private String parserImpl;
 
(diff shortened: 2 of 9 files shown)


lewismc:
Confirmed this PR addresses the [security hotspots](https://sonarcloud.io/project/security_hotspots?id=apache_nutch&pullRequest=904) and passes tests.

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
```diff
@@ -93,6 +86,82 @@ public class HtmlParser implements Parser {
...
+    return end > start ? s.substring(start, end) : null;
+  }
+
+  /**
+   * Finds charset from HTML string using linear scans only (no backtracking regex).
+   * Checks meta http-equiv Content-Type then HTML5 meta charset.
+   * Package-private for unit testing.
+   */
+  static String extractCharsetFromMeta(String str) {
+    String lower = str.toLowerCase();
```
sebastian-nagel, line 133: Should add `Locale.ROOT`.

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
```diff
@@ -102,20 +171,7 @@ private static String sniffCharacterEncoding(byte[] content) {
     // {U+0041, U+0082, U+00B7}.
     String str = new String(content, 0, length, StandardCharsets.US_ASCII);
```
sebastian-nagel, line 172: The content bytes are converted into a String assuming ASCII encoding. If a "hand-crafted" scanner is used, could directly operate on the bytes avoiding the conversion to a string. Alternatively, could use a CharSequence wrapping the bytes, cf. [ByteArrayCharSequence](https://github.com/commoncrawl/nutch/blob/cc/src/java/org/commoncrawl/util/ByteArrayCharSequence.java). But this can be an improvement for later.

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
```diff
@@ -93,6 +86,82 @@ public class HtmlParser implements Parser {
...
+    while (true) {
+      int metaStart = lower.indexOf(META_TAG_START, pos);
+      if (metaStart < 0) {
+        break;
+      }
+      int tagEnd = str.indexOf('>', metaStart);
+      if (tagEnd < 0) {
+        break;
+      }
+      String tagContent = str.substring(metaStart, tagEnd);
```
sebastian-nagel, line 144: Assuming that indexes in the lower-case string and the original one are the same might be dangerous, for example, in German there is the pair "ß" <> "SS" (would held for uppercasing the string). See [Unicode TR #21 Case Mappings](https://www.unicode.org/L2/L1999/99190.htm): "Case mappings may produce strings of different length than the original."

sebastian-nagel: Of course, if we strictly ensure that the method is used on String only including ASCII characters and using always the root locale, this should be safe.

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
```diff
@@ -93,6 +86,82 @@ public class HtmlParser implements Parser {
...
+      String tagLower = tagContent.toLowerCase();
```
sebastian-nagel, line 145: Same: must use `Locale.ROOT`.

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
```diff
@@ -64,15 +62,10 @@ public class HtmlParser implements Parser {
...
-  private static Pattern metaPattern = Pattern.compile(
-      "<meta\\s+([^>]*http-equiv=(\"|')?content-type(\"|')?[^>]*)>",
-      Pattern.CASE_INSENSITIVE);
-  private static Pattern charsetPattern = Pattern.compile(
-      "charset=\\s*([a-z][_\\-0-9a-z]*)", Pattern.CASE_INSENSITIVE);
-  private static Pattern charsetPatternHTML5 = Pattern.compile(
-      "<meta\\s+charset\\s*=\\s*[\"']?([a-z][_\\-0-9a-z]*)[^>]*>",
-      Pattern.CASE_INSENSITIVE);
+  private static final String META_TAG_START = "<meta";
+  private static final String CHARSET_EQ = "charset=";
```
sebastian-nagel, line 66: The regex allows white space between "charset" and "=".

src/java/org/apache/nutch/parse/ParseOutputFormat.java
```diff
@@ -177,8 +194,8 @@ public RecordWriter<Text, Parse> getRecordWriter(TaskAttemptContext context)
     Path data = new Path(new Path(out, ParseData.DIR_NAME), name);
     Path crawl = new Path(new Path(out, CrawlDatum.PARSE_DIR_NAME), name);
 
-    final String[] parseMDtoCrawlDB = conf.get("db.parsemeta.to.crawldb", "")
```
sebastian-nagel, line 180: It's a string from the configuration, it's controlled by the user and should be short. It's parsed once, so this is for sure not critical.

Apart from that should simply rely on [getTrimmedStrings](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/conf/Configuration.html#getTrimmedStrings(java.lang.String,java.lang.String...)).

src/plugin/parse-html/src/java/org/apache/nutch/parse/html/HtmlParser.java
```diff
@@ -93,6 +86,82 @@ public class HtmlParser implements Parser {
...
+   * Extracts charset value from a string like "charset=utf-8" or "charset = utf-8".
+   * Uses linear scan to avoid ReDoS. Value must start with [a-z] and contain only [a-z0-9_-].
+   */
+  private static String extractCharsetValue(String s, int fromIndex) {
+    int idx = s.indexOf(CHARSET_EQ, fromIndex);
+    if (idx < 0) {
+      return null;
+    }
+    int start = idx + CHARSET_EQ.length();
+    while (start < s.length() && (s.charAt(start) == ' ' || s.charAt(start) == '\t')) {
```
sebastian-nagel, line 99: `\\s` matches more characters than blank (U+0020) and the tab character.

sebastian-nagel (changes requested):
Hi @lewismc, thanks!

So far, this is only a partial review. See the inline comments. Without having tried it, I expect some regressions, e.g. when extracting the charset from `<meta charset    = "UTF-8" />`. Replacing regular expressions by hand-written scanner code is difficult and makes code maintenance much more difficult. Maybe use a scanner generator instead, e.g. Ragel? Alternatively, we could use [RE2/J](https://github.com/google/re2j) or [dk.brics.automaton](https://www.brics.dk/automaton/) (see urlfilter-automaton) which are faster (and more safe) because they do not backtrack.

In addition, the URL validator needs a more substantial fix, see NUTCH-2986.

lewismc:
Thank you for the sincere review @sebastian-nagel I appreciate it. This PR was too adventurous and too large. I'll close it and move to these issues later on.

PR #908, opened 2026-04-12
NUTCH-3165 Remove the Nutch web service

lewismc:
PR for [NUTCH-3165](https://issues.apache.org/jira/browse/NUTCH-3165). As mentioned in the Jira ticket, this PR implements the nuclear option of outright removing the entire Nutch service. 
Any future service can/should be designed and implemented using the OpenAPI spec defined in [NUTCH-2932](https://issues.apache.org/jira/browse/NUTCH-2932).


lewismc:
TestFetcher was breaking due a clash between the transitive Jetty 9 dependency defined by Hadoop 3.4.3 and the explicit dependency defined in `ivy.xml`.
I've used [MockServer](https://www.mock-server.com/) before with success so decided to remove the clash by replacing the Jetty 12 dependency with MockServer. I hope this prevents future clashes when we upgrade Hadoop.

sebastian-nagel (approved):
+1 Thanks, @lewismc!

PR #909, opened 2026-04-13
NUTCH-3168 Sandbox Commons JEXL usage in crawl and index pipelines

lewismc:
PR for [NUTCH-3168](https://issues.apache.org/jira/browse/NUTCH-3168).
Some unit tests validate basic/common Jexl expressions and some edge cases.
I decided to implement a brand new configuration property `nutch.jexl.disable.sandbox` which allows users in trusted environments to bypass the sandbox. Another important design consideration was to maintain compatibility with ALL existing Jexl scripts users may have written prior to this patch.
Finally, Jexl expression handling is standardized across pipelines so behavior is predictable and auditable.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -1606,7 +1606,10 @@
   <value>plugins</value>
   <description>Directories where Nutch plugins are located.  Each
   element may be a relative or absolute path.  If absolute, it is used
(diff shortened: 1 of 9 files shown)


sebastian-nagel (approved):
+1 Looks good. Thanks, @lewismc!

Tested the CrawlDbReader with a handful JEXL expressions to filter the dumped CrawlDb.

PR #912, opened 2026-05-07
NUTCH-3174 protocol-okhttp: request may hang despite http.time.limit is set

sebastian-nagel:
- set OkHttp's call timeout to the value of http.time.limit (if not -1)
- add check whether http.time.limit is longer than http.timeout
- more verbose logging


lewismc (approved):
OK nice catch. I tested the logging statement with some triggering values for `http.timeout` and `http.time.limit`. 

+1

sebastian-nagel:
Thanks for review and testing, @lewismc!

PR #915, opened 2026-05-13
NUTCH-3177 Fetcher to report idle threads not as hung threads

sebastian-nagel:
- reset reprUrl in FetcherThread after fetch is finished
- report idle threads properly


lewismc (approved):
+1 good distinction and logging improvement. 
I wonder if the `hungThreadsCounter` should be renamed to accommodate idle threads as well?

sebastian-nagel:
> I wonder if the `hungThreadsCounter` should be renamed to accommodate idle threads as well?

Thanks, good point. Or, it should not count the idle threads in. Or even better, exit earlier if there are only queues, all with delays exceeding the hard timeout induced by MapReduce. I'll need to think about it. Will open a new issue.

PR #918, opened 2026-05-23
NUTCH-3164 Catch specific exceptions in CrawlDbFilter so plugin errors no longer silently drop URLs

prakharchaube:
## Summary

  Both catch blocks in `CrawlDbFilter.map()` caught generic `Exception` and set `url = null`, which silently dropped URLs both for legitimate filtering reasons and for plugin
  programming errors (NPE, etc.). The latter masked plugin bugs as ordinary filtering decisions.
  
  ## Changes

  **Normalizer block**
  - `MalformedURLException` — the only legitimate reason to drop. Tracked via `ErrorTracker` (`ErrorType.URL`) and no longer increments `urlsFilteredCounter`, which conflated
  filtering with malformed input.
  - `RuntimeException` — logged at ERROR and tracked, URL is **not** dropped so plugin bugs do not silently delete data.
  
  **Filter block**
  - `URLFilterException` — per the `URLFilter` contract, reserved for internal filter failures (rejection is signaled by returning `null`). Logged at ERROR and tracked, URL is
  **not** dropped.
  - `RuntimeException` — same handling as above.
  
  All error paths now use `ErrorTracker` for categorized counters and log at ERROR rather than WARN, per [@lewismc's recommendation in the JIRA 
  discussion](https://issues.apache.org/jira/browse/NUTCH-3164).
  
  ## JIRA

  [NUTCH-3164](https://issues.apache.org/jira/browse/NUTCH-3164)
  
  ## Test plan

  - [x] `ant compile` passes locally
  - [ ] No new unit tests in this PR; happy to add one if reviewers want a mock plugin throwing each exception type (flag as follow-up otherwise)
  
  ## Out of scope (separate tickets if desired)
  
  - Same `catch (Exception)` pattern exists in `Injector.filterNormalize` and ~20 other call sites of `normalizers.normalize` / `filters.filter` — per lewismc comment, those
  should be rolled out separately.
  - Whether to track normalized URLs as a metric system-wide.

### DIFF
--- a/src/java/org/apache/nutch/crawl/CrawlDbFilter.java
+++ b/src/java/org/apache/nutch/crawl/CrawlDbFilter.java
@@ -18,6 +18,7 @@ package org.apache.nutch.crawl;
 
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
+import java.net.MalformedURLException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -25,7 +26,9 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapreduce.Counter;
 import org.apache.hadoop.mapreduce.Mapper;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.nutch.metrics.ErrorTracker;
 import org.apache.nutch.metrics.NutchMetrics;
+import org.apache.nutch.net.URLFilterException;
 import org.apache.nutch.net.URLFilters;
 import org.apache.nutch.net.URLNormalizers;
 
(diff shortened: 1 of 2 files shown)


sebastian-nagel (commented):
Thanks, @prakharchaube, looks great!

I've tried hard to raise in one of the existing URL filters and normalizers an exception which is uncaught and not converted into a MalformedURLException.

This makes testing it hard. But I think it's definitely no blocker. The code is clear.

If you want to write also a unit test, let us know. Otherwise I've merge this PR next week. Thanks again!

sebastian-nagel (approved):

prakharchaube:
Hi @sebastian-nagel I have added a UT file, let me know if anything else is needed!

sebastian-nagel:
Excellent, @prakharchaube!

PR #919, opened 2026-05-25
NUTCH-3173 protocol-okhttp: store OkHttp's internal URL in response metadata

lfoppiano:
This PR solves https://issues.apache.org/jira/browse/NUTCH-3173. 

One example URL https://www.h.country

### DIFF
--- a/src/plugin/lib-http/src/java/org/apache/nutch/protocol/http/api/HttpRobotRulesParser.java
+++ b/src/plugin/lib-http/src/java/org/apache/nutch/protocol/http/api/HttpRobotRulesParser.java
@@ -170,7 +170,8 @@ public class HttpRobotRulesParser extends RobotRulesParser {
             LOG.debug("Following robots.txt redirect: {} -> {}", robotsUrlRedir,
                 redirectionLocation);
             try {
-              robotsUrlRedir = new URL(robotsUrlRedir, redirectionLocation);
+              robotsUrlRedir = ((HttpBase) http).resolveUrl(robotsUrlRedir,
+                  redirectionLocation);
             } catch (MalformedURLException e) {
               LOG.info(
                   "Failed to resolve redirect location for robots.txt: {} -> {} ({})",
--- a/src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpResponse.java
+++ b/src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpResponse.java
@@ -121,6 +121,17 @@ public class HttpResponse implements Response {
       client.getParams().setParameter("http.useragent", http.getUserAgent()); // NUTCH-1941
       code = client.executeMethod(get);
 
+      // When followRedirects=true HC3 walks the redirect chain internally;
+      // getURI() returns the final URI. Capture it so getUrl() honors the
+      // contract — without this, robots.txt redirects via this plugin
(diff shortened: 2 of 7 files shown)


lewismc (approved):
+1 I tested this locally and it works. A unit test(s) would be great but not essential. I'm approving. Thank you @lfoppiano

src/plugin/lib-http/src/java/org/apache/nutch/protocol/http/api/HttpRobotRulesParser.java
```diff
@@ -170,7 +170,8 @@ public BaseRobotRules getRobotRulesSet(Protocol http, URL url,
             LOG.debug("Following robots.txt redirect: {} -> {}", robotsUrlRedir,
                 redirectionLocation);
             try {
-              robotsUrlRedir = new URL(robotsUrlRedir, redirectionLocation);
+              robotsUrlRedir = ((HttpBase) http).resolveUrl(robotsUrlRedir,
```
sebastian-nagel, line 173: Since the Protocol class provides a default implementation (a fall-back to the Java URL class), there should be no cast to HttpBase necessary to call the overridden method.

src/plugin/protocol-httpclient/src/java/org/apache/nutch/protocol/httpclient/HttpResponse.java
```diff
@@ -121,6 +121,17 @@ public class HttpResponse implements Response {
       client.getParams().setParameter("http.useragent", http.getUserAgent()); // NUTCH-1941
       code = client.executeMethod(get);
 
+      // When followRedirects=true HC3 walks the redirect chain internally;
+      // getURI() returns the final URI. Capture it so getUrl() honors the
+      // contract — without this, robots.txt redirects via this plugin
+      // report the original URL even though a different URL was fetched.
+      try {
+        this.url = new URL(get.getURI().toString());
```
sebastian-nagel, line 129: Good catch. Of course, it would also affect other fetches, not only the robots.txt - although the robots.txt fetch is the only occurrence where `getResponse` is called with `followRedirects = true`.

Some protocol implementations, such as protocol-http or protocol-okhttp, ignore the parameter `followRedirects` anyway, and delegate the control over redirects to the caller.

src/plugin/protocol-okhttp/src/java/org/apache/nutch/protocol/okhttp/OkHttpResponse.java
```diff
@@ -102,7 +102,14 @@ public OkHttpResponse(OkHttp okhttp, URL url, CrawlDatum datum)
     }
 
     Request request = rb.build();
-    okhttp3.Call call = okhttp.getClient(url).newCall(request);
+
+    // OkHttp parsed the URL via HttpUrl; that's the form actually going on
+    // the wire (IDN→punycode, repeated-slash repair, host lowercasing).
+    this.url = request.url().url();
+    if (LOG.isDebugEnabled() && !this.url.toString().equals(url.toString())) {
```
sebastian-nagel, line 109: Good not to call [URL.equals(url)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URL.html#equals(java.lang.Object)) which would trigger a DNS lookup.

sebastian-nagel (approved):
Thanks, @lfoppiano!

> A unit test(s) would be great but not essential.

I'd second this. Can be added later.

PR #920, opened 2026-05-27
Add AGENTS.md + SECURITY.md linking the project's security model

potiuk:
**This is a proposal for the PMC to review — please correct, reject, or discuss as needed.** Nothing here is a requirement; the maintainer is the decision-maker.

## What this PR does

Adds two new files at the repo root:

- **`AGENTS.md`** — agent-instructions file with a Security section linking to `SECURITY.md`. The conventional first stop for automated agents (security scanners, code analyzers, AI assistants) operating on this repository.
- **`SECURITY.md`** — short pointer to the canonical security documents on the project website at <https://nutch.apache.org/documentation/security/>, plus the standard ASF reporting flow (`security@apache.org`). No threat-model content is duplicated — the website stays the source of truth.

Together they make the conventional `AGENTS.md → SECURITY.md → website` discoverability chain mechanically followable.

## Why

Two practical drivers:

1. **GitHub UI affordance.** GitHub's "Report a vulnerability" button surfaces the contents of `SECURITY.md` at the repo root. Without one, well-meaning reporters file public issues against perceived security gaps. The new `SECURITY.md` redirects them to the website's threat-model + reporting flow.

2. **Agent-driven security tooling discovery.** An automated agentic security scan the ASF Security team is piloting needs to mechanically locate the project's threat model via the conventional `AGENTS.md → SECURITY.md` chain at the designated commit. Without that chain resolving, the scan refuses to run (refusing upfront beats wasting reviewer cycles on a noise-heavy run against an unknown model).

The PMC opted in on a private thread with the ASF Security team; Sebastian Nagel picked option (b) — Security team opens the PR — in his 2026-05-26 reply.

## What this PR does NOT do

- It does **not** alter the existing security model. The model at <https://nutch.apache.org/documentation/security/#security-model> stays the source of truth.
- It does **not** introduce a new reporting alias. Reports continue to flow through `security@apache.org`.
- It does **not** touch other repository files.

Per Sebastian's "template or incomplete version, we can complete during PR review" note: this is exactly that — minimal scaffolding that lets the discoverability chain pass mechanically. The PMC can extend `AGENTS.md` over time with project-specific guidance for AI coding agents (general contribution rules, build conventions, etc.) — it's the agent-instruction file the security pointer lives in.

Questions / pushback / scope expansion welcome.

PR #922, opened 2026-06-06
Add in-repo threat model and point AGENTS.md/SECURITY.md at it

potiuk:
**This is a v0 draft proposal for the Nutch PMC to review — please correct, reject, or discuss as needed.** Following up on Lewis's note to go ahead and draft the model so it keeps momentum.

**Context.** The ASF Security team is preparing the project for an automated agentic security scan we're piloting; the scan runs against a threat model so its output is signal rather than noise. Discoverability already landed in #920; this PR adds the model content.

**What's in this PR:**
- **`THREAT_MODEL.md`** (new) — a v0 threat model written from your website security model + the codebase, following the [threat-model-producer rubric](https://gist.github.com/potiuk/da14a826283038ddfe38cc9fe6310573). It is a **strict superset of the website security model** — nothing there is dropped; the sections the website page didn't cover (adversary model, enumerated properties, known non-findings, triage dispositions) are added and tagged *(inferred)* for you to confirm / correct / strike. Draft confidence ~10 documented / 22 inferred.
- **`AGENTS.md` + `SECURITY.md`** — re-pointed so the chain resolves `AGENTS.md → SECURITY.md → THREAT_MODEL.md`, keeping the website references intact.

**The framing to sanity-check:** Nutch fetches + parses untrusted web content by design, so crawler "SSRF" (reaching internal/arbitrary URLs) and "parses hostile HTML/XML" are by-design — scoped by your URL filters, not by Nutch refusing. The in-model adversary is the **malicious crawled-content supplier** (XXE, parser DoS, decompression bombs, ReDoS in URL-filter regex). §11a captures those recurring false positives.

**What we'd need from the PMC:** walk §14 (3 waves) — a one-line confirm / correct / strike per question is enough; wave 1 (trusted-environment posture incl. `nutch-server`, crawler-SSRF/scope, the crawled-content adversary) shapes the rest. §14.7 asks whether this in-repo model or the website page should be canonical.

If you'd rather adjust the approach, comment on the PR or close it — entirely your call.

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -1,0 +1,327 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to you under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Apache Nutch — Security Threat Model
+
+## §1 Header
+
+- **Project:** Apache Nutch (core crawler — `apache/nutch`)
+- **Modeled against:** `master` HEAD as of 2026-06-06.
+- **Authors:** ASF Security team (v0 draft, generated via the
+  `threat-model-producer` rubric at the Nutch PMC's request), for the PMC to review.
+- **Status:** **DRAFT v0 — draft-first, not yet maintainer-ratified.** Built as a
+  superset of the project's existing website security model; the sections that
+  were not covered there are *(inferred)* and must be confirmed (see §14).
+- **Version binding:** versioned with the project; a report against release *N*
+  is triaged against the model as it stood at *N*.
+- **Reporting cross-reference:** findings that violate a §8 property should be
+  reported privately per the project's disclosure channel (`security@apache.org`);
+  findings under §3 or §9 are closed citing this document.
+- **Provenance legend:** *(documented)* = stated in Nutch's docs/website security
+  model; *(maintainer)* = confirmed by a Nutch PMC member; *(inferred)* = reasoned
+  from docs/code/domain knowledge, not yet confirmed (each has a §14 question).
+- **Draft confidence:** ~10 documented / 0 maintainer / ~22 inferred.
+- **Relationship to the website model:** this document is a strict superset of
+  the security model published at
+  <https://nutch.apache.org/documentation/security/#security-model>; nothing that
+  page asserts is dropped or weakened here. Where this draft adds a section the
+  page did not cover (adversary model, enumerated properties, known non-findings,
+  triage dispositions), it is tagged *(inferred)* pending PMC confirmation.
+
+**What Nutch is.** Apache Nutch is an extensible, Hadoop-based open-source web
+crawler: an operator runs it (in local or distributed/batch mode, or via the
+`nutch-server` REST control API) to fetch, parse, and index web content at
+scale through a plugin architecture (protocol / parse / index / scoring
+plugins; parsing largely via Apache Tika). The defining security fact is that
+**Nutch fetches and parses content from the open, untrusted web by design** —
+the crawled bytes are attacker-controllable input, and the threat model is about
+robustly handling that input and about which controls are the operator's job,
+not about preventing Nutch from reaching or parsing hostile content.
+
+## §2 Scope and intended use
+
+- **Primary intended use** *(documented)*: an operator-deployed crawler run
+  **in a trusted environment**, fetching web content into a crawl store and
+  handing parsed content to an indexer. The website model states Nutch is
+  "designed to operate in trusted environments, either locally or on a Hadoop
+  cluster."
+- **Deployment shape** *(documented/inferred)*: batch crawl jobs (local or on
+  Hadoop), plus an optional `nutch-server` REST API for orchestration.
+- **Caller roles:**
+  - **operator / admin** — trusted; owns seeds, URL filters, plugin config, the
+    Hadoop/host environment, and the `nutch-server` endpoint.
+  - **crawled-content supplier** — the untrusted web: every fetched page,
+    redirect, `robots.txt`, sitemap, feed, and embedded resource is
+    attacker-controllable input. *(inferred — the core in-model adversary.)*
+  - **REST API client** — in the trusted environment per the website model; the
+    legacy REST API provided no authentication. *(documented)*
+
+**Component-family table** *(inferred — confirm in §14)*:
+
+| Family | Entry point | Touches outside process? | In model? |
+| --- | --- | --- | --- |
+| Fetcher / protocol plugins | `protocol-http(client)`, etc. | network (arbitrary URLs) | **yes** — consumes untrusted content |
+| Parser plugins (Tika, HTML, feed, etc.) | `parse-*` | CPU/memory on untrusted bytes | **yes** |
+| URL filtering / normalization | `urlfilter-*`, `urlnormalizer-*` (regex) | — | **yes** (scoping boundary) |
+| Crawl store / DB | local FS / HDFS | disk | **yes** |
+| Indexer plugins | `indexer-solr`, `-elastic`, … | network (backend) | **boundary** — backend is the operator's |
+| `nutch-server` REST API | HTTP control endpoint | network | **yes** |
+| Bundled/contrib plugins, examples | various | varies | **per-plugin** — confirm supported set (§14) |
+
+## §3 Out of scope (explicit non-goals)
+
+- **Defending a Nutch deployment exposed outside a trusted environment** (e.g.
+  an internet-reachable `nutch-server` with no fronting auth). The website model
+  scopes Nutch to trusted environments. *(documented)*
+- **Preventing Nutch from fetching arbitrary or internal URLs.** Reaching URLs
+  is the crawler's function; restricting *which* URLs is the operator's job via
+  URL filters / seed scoping (so crawler "SSRF" is operator-config, see §9/§11a).
+  *(inferred)*
+- **The security of indexer/storage backends** (Solr/Elasticsearch/HDFS) and the
+  Hadoop cluster — Nutch writes to them; it does not own their security.
+  *(inferred)*
+- **Contrib / unsupported plugins / examples** — threat-modeled separately;
+  confirm the supported plugin set in §14. *(inferred)*
+
+## §4 Trust boundaries and data flow
+
+Two boundaries matter, and they are different from a typical service:
+1. **The fetch boundary (primary).** Everything Nutch fetches from the web is
+   untrusted and crosses into the parser/store. The trust transition is
+   "bytes-from-the-internet → parsed structures." *(inferred)*
+2. **The operator/config boundary.** Seeds, URL filters, plugin selection, and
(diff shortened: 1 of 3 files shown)


lewismc:
Thank you again @potiuk , the CI failure relates to markdownlint which we can fix easily. 
I will study this PR in detail tomorrow. Thank you,

THREAT_MODEL.md
```diff
@@ -0,0 +1,327 @@
...
+  parser libraries over untrusted bytes; with `nutch-server`, opens a listening
+  HTTP port. Not expected to run as root.
+
+## §5a Build-time and configuration variants
+
+Config knobs that move the security envelope *(inferred — confirm in §14)*:
+
+| Knob | Default | Effect | Maintainer stance |
+| --- | --- | --- | --- |
+| `nutch-server` REST API | not started unless launched; **no auth** | An exposed control endpoint with no auth on an untrusted network = unauthenticated control | **?** trusted-env-only posture — §14.1 |
```
sebastian-nagel, line 141: Maybe we should relax the Yetus rules for the line length, at least, for table rows? I'm not aware of a way to have a table row spread over multiple lines. Except using HTML tables.

THREAT_MODEL.md
```diff
@@ -0,0 +1,327 @@
...
+| `KNOWN-NON-FINDING` | Matches a §11a pattern. | §11a |
+| `MODEL-GAP` | Cannot be routed above → revise the model. | §12 |
+
+## §14 Open questions for the maintainers
+
+Grouped in waves; each states a **proposed answer** to confirm/correct/strike.
+Every *(inferred)* tag above maps to one of these.
+
+**Wave 1 — scope & adversary (these shape everything):**
+1. **Trusted-environment posture / `nutch-server`.** Proposed: the supported
```
sebastian-nagel, line 298: Yes, `nutch-server` must run in a trusted environment. I will be removed in 1.23, resp. is already removed in the master branch.

THREAT_MODEL.md
```diff
@@ -0,0 +1,327 @@
...
+   posture is "trusted environment only"; an exposed no-auth `nutch-server` is
+   `OUT-OF-MODEL: non-default-build`. Correct? (→ §5a, §3, §11a)
+2. **Crawler SSRF / scope.** Proposed: fetching internal/arbitrary URLs is
```
sebastian-nagel, line 301: Yes, correct.

THREAT_MODEL.md
```diff
@@ -0,0 +1,327 @@
...
+   by-design and controlled by operator URL filters, not a Nutch vulnerability;
+   only *escaping the configured scope* is in-model. Correct? (→ §9, §11a)
+3. **Primary adversary = the crawled-content supplier.** Proposed: the main
```
sebastian-nagel, line 304: Yes, agreed.

In addition, we must not trust and assume that these can be controlled by an attacker:
- DNS, e.g., an attacker might point the DNS record for `attackers-domain.xyz` to an IP address not controlled by the attacker, including internal (local, loopback) IP addresses. IP address filters are available for the `protocol-okhttp` plugin. The filters need to be configured accordingly.
- HTTP headers, including the "Location" header for redirects.
- Every message in the protocol stack, including TCP/IP, TLS handshakes, etc. Although Nutch relies on Java core or third-party libraries for these parts.

THREAT_MODEL.md
```diff
@@ -0,0 +1,327 @@
...
+   redirects), and parser robustness against it is the core property. Agree? Any
+   other in-model adversary? (→ §7)
+
+**Wave 2 — properties & parsers:**
+4. **Parser hardening (XXE / bombs).** Are XML/feed parsers configured against
+   XXE and decompression/entity bombs by default, or is that operator config?
+   (→ §8, §9)
+5. **Resource line.** Where is the line between an in-model parser-DoS on crafted
+   content and an out-of-model "expensive legitimate crawl"? (→ §8)
+6. **Supported plugin set.** Which `protocol-*` / `parse-*` / `indexer-*` plugins
```
sebastian-nagel, line 315: Yes.

sebastian-nagel (approved):
Thanks, @potiuk. Looks good.

See also the inline comments.

@lewismc, let me know who shall polish the PR to final version.

THREAT_MODEL.md
lewismc: Aye I think we should relax that. I'll investigate. I also want to relax the SonarCloud >80% test coverage per PR. That is unrealistic. I'll address both.

THREAT_MODEL.md
lewismc: Motigated. Resolving.

THREAT_MODEL.md
lewismc: I'd support augmenting the model with this exact narrative.

lewismc:
The Yetus markdownlint issues can be relaxed and addressed elsewhere.

lewismc:
@sebastian-nagel
Is there attention to be paid to [§8.1](https://github.com/potiuk/nutch/blob/asf-security/threat-model-2026-06-06/THREAT_MODEL.md#8-security-properties-the-project-provides) the `fetcher.parse` and `fetcher.store.content` properties (although mitigated by default) can also result in corrupted segments in error scenarios. Maybe this is covered within the threat model language but thought I'd double check.

I also think that §8.1 **No execution of fetched content** needs more thought. I'm fairly sure CVE's have been registered against Tika for code execution from archive files. I may be wrong but thought I'd raise it as well. **EDIT** I just noticed this is covered by the **Decompression/zip bombs** stated in §9

Here's my take on §14

> Wave 1 — scope & adversary (these shape everything):
Trusted-environment posture / nutch-server. Proposed: the supported posture is "trusted environment only"; an exposed no-auth nutch-server is OUT-OF-MODEL: non-default-build. Correct? (→ §5a, §3, §11a)

Correct

> Crawler SSRF / scope. Proposed: fetching internal/arbitrary URLs is by-design and controlled by operator URL filters, not a Nutch vulnerability; only escaping the configured scope is in-model. Correct? (→ §9, §11a)

Correct

> Primary adversary = the crawled-content supplier. Proposed: the main in-model attacker is whoever controls fetched content (hostile HTML/XML/feeds/ redirects), and parser robustness against it is the core property. Agree? Any other in-model adversary? (→ §7)

Agree. No further comments, for now.

> Wave 2 — properties & parsers: 4. Parser hardening (XXE / bombs). Are XML/feed parsers configured against XXE and decompression/entity bombs by default, or is that operator config? (→ §8, §9) 

By default via http.content.limit and CVE fixes in tika parsers. What about parse-zip?

> 5. Resource line. Where is the line between an in-model parser-DoS on crafted content and an out-of-model "expensive legitimate crawl"? (→ §8)

This is operator dependent and would need to be assessed on a crawl-by-crawl basis. Instrumentation efforts may improve this understanding and we do have [ErrorTracker](https://github.com/apache/nutch/blob/master/src/java/org/apache/nutch/metrics/ErrorTracker.java) to assist operators. 

6. Supported plugin set. Which protocol-* / parse-* / indexer-* plugins are first-class for security vs. contrib/unsupported? (→ §2/§3/§5a)

The Nutch project has not concept of `contrib`. I believe the project considers all _official_ plugins falling into the first class bucket. Any other plugins are out with the project's control. 

> Wave 3 — meta: 7. Canonicalization. This in-repo THREAT_MODEL.md is drafted as a superset of the website security-model page. Proposed: SECURITY.md points here for the full model, and the website page stays the operator-facing how-to. Agree, or should the website page remain canonical with this as a supplement? (→ meta)

Agree, we can do that easily.

THREAT_MODEL.md
sebastian-nagel: Ok. I'll update the PR. Need to resolve the merge conflict anyway.

sebastian-nagel:
> Is there attention to be paid to [§8.1](https://github.com/potiuk/nutch/blob/asf-security/threat-model-2026-06-06/THREAT_MODEL.md#8-security-properties-the-project-provides) the `fetcher.parse` and `fetcher.store.content` properties (although mitigated by default) can also result in corrupted segments in error scenarios.

- `fetcher.parse`: The fetcher uses the ParseUtil class to let the parser run in an ExecutorService with a default timeout of 30 seconds. In this regard, there is no difference to parsing the segments content in a separate job. Of course, a parsing fetcher map task may require more resources (mostly RAM, less CPU) than a parser map task because it does both fetching and parsing, and also because of its multi-threaded implementation.
- `fetcher.store.content`: By default, `http.content.limit` (resp. `ftp.content.limit`) restricts the size of stored content to 1 MiB per document. To run a separate parse job, content needs to be stored in the segment. Truncated contents, exceeding the content limit, are not parsed by default (property `parser.skip.truncated`).

In case of an unhandled error, the fetch or parse task should fail, and not produce any final output. That is, the segment is incomplete (no fetch or parse data), but not "corrupted".

Generally, we might add notes about resource requirements and limitations:

1. Nutch's Fetcher is multi-threaded in nature. It buffers the fetched content before it's spilled to disk, caches robots.txt rules, and optionally parses the fetched documents. The Fetcher requires sufficient RAM to hold data temporarilly in memory and sufficient CPU to process the content in parallel threads. The resource requirements depend on the configuration and the number of assigned fetcher threads.

2. Nutch's processing "pipeline" to fetch, parse and index web content is designed to process reasonable-size web documents. By default, the size of a single document (HTML, PDF, or any other document format) is limited to 1 MiB. Nutch can process larger document, if sufficient RAM is configured. However, there are limitations:
   - Nutch buffers documents using a Java byte array which is limited to about 2 GiB. This is a hard limitation.
   - Nutch cannot process "indefinite" content streams (internet radio or TV). In addition to the content limit, there are strict timeouts to download a single document.
   - Usually, it's already troublesome to make Nutch processing documents exceeding 100 MiB. In other words, Nutch isn't the ideal tool to collect, e.g., very large PDF files.

Where we should add these limitations? It's also a description what Nutch cannot be used for.


> What about parse-zip?

It's rarely used, and a simple plugin, parsing zip files recursively. Likely, it's not safe. Again, the 30 second parse timeout protects it from the worst outcome. But we very eventually need to drop it, fix it or mark it as unsafe, that is to be used only for trusted content.

Btw., zip bombs also threaten the protocol plugins per `Content-Encoding` and `Transfer-Encoding`, at least, for the encodings `deflate`, `gzip` and `br`. Since the content limit is applied to the decoded/uncompressed content, the protocol plugins should be safe. But we eventually should test all protocol plugins.

>> 5. Resource line. Where is the line between an in-model parser-DoS on crafted content and an out-of-model "expensive legitimate crawl"? (→ §8)

> This is operator dependent and would need to be assessed on a crawl-by-crawl basis.

+1

>> Wave 3 — meta: 7. Canonicalization. This in-repo THREAT_MODEL.md is drafted as a superset of the website security-model page. Proposed: SECURITY.md points here for the full model, and the website page stays the operator-facing how-to. Agree, or should the website page remain canonical with this as a supplement? (→ meta)

> Agree, we can do that easily.

+1

sebastian-nagel:
> **What we'd need from the PMC:** walk §14 (3 waves) — a one-line confirm / correct / strike per question is enough;

@potiuk, just for clarification: are the comments in the PR sufficient, or shall we put the answers into the markdown? If in the markdown, shall it be a pair of question and answer, or shall we rephrase every point, so that we get a statement.

potiuk:
All fine. I will take it from here.

potiuk:
Thanks Sebastian — and thanks for the approve.

PR comments are completely sufficient on our side: a one-line confirm / correct / strike per §14 question, wherever it's easiest for you to write it (inline on the PR is perfect). You don't need to restructure anything.

For the durable record, we'll fold the settled answers back into `THREAT_MODEL.md` ourselves as **declarative statements** (not a Q&A list) in the final pass — so the document reads as the project's own model rather than an interview transcript. So: you answer however is lowest-effort; we do the markdown cleanup.

On the `yetus` markdownlint failure — agreed it's cosmetic; we'll fix the lint in our cleanup commit so the PR goes green before merge.

potiuk:
Thanks for approving! Ready to merge whenever convenient — the one red check is an unrelated/flaky CI job on a docs-only change.

sebastian-nagel:
Thanks, @potiuk! Merging...

PR #923, opened 2026-06-08
NUTCH-3180 BasicURLNormalizer missing catching ICUInputTooLongException 

lfoppiano:
This PR fixes the issue https://issues.apache.org/jira/browse/NUTCH-3180. 
It hardens Nutch’s IDNA2008 hostname conversion so that unchecked ICU exceptions (e.g., from UTS46/Punycode) are converted into MalformedURLException, allowing callers like BasicURLNormalizer/URLCleaner to reject bad URLs instead of crashing a task.

Changes:
  -  Wrap ICU IDNA.nameToASCII/nameToUnicode in URLUtil.convertIDNA2008 with a try/catch to translate unchecked ICU/UTS46/Punycode exceptions into MalformedURLException (with cause attached).
  -  Add regression tests intended to cover possible Exception thrown by the underlying library


sebastian-nagel (approved):
Thanks, @lfoppiano! Precise fix and concise unit tests.

lfoppiano:
@sebastian-nagel should I fix the issue with sonaqubecloud? I'm not sure what would be the preferred way to fix it.

sebastian-nagel:
> should I fix the issue with sonaqubecloud?
>> "Either log this exception and handle it, or rethrow it with some contextual information."

You might remove the debug log - it's on you. The original exception is preserved per `initCause(e)`, and most callers log the MalformedURLException anyway. But this is really a minor issue - there are many occurrences in the Nutch codebase where an error is logged and then thrown as exception.

lfoppiano:
OK. Removed the log. I though it would be useful for this specific issue, but indeed it's an antipattern 😄

sebastian-nagel:
> Removed the log.

@lfoppiano: there is a second logging statement, it's still in the code. Do you want to remove it?

lfoppiano:
@sebastian-nagel sorry I forgot it. Now the second log is also gone.

PR #927, opened 2026-06-12
NUTCH-3184 Upgrade OkHttp to 3.4.0

sebastian-nagel:


lewismc (approved):
+1 @sebastian-nagel 
I wonder if we can make use of the [new superpowered interceptors](https://square.github.io/okhttp/changelogs/changelog/#version-540)?

sebastian-nagel:
> I wonder if we can make use of the [new superpowered interceptors](https://square.github.io/okhttp/changelogs/changelog/#version-540)?

Yes, I've also stumbled about this notice in the Changelog, but could resist to try them. No hurry. It's also important to follow what our colleagues at StormCrawler do with OkHttp. Recently, they even uncovered a bug in Nutch's pluging, see #929 / NUTCH-3187.

PR #928, opened 2026-06-12
NUTCH-3185 Upgrade NekoHTML to 1.9.22

sebastian-nagel:
- Upgrade the Neko version in lib-nekohtml (used by parse-html and headings plugins)
- Remove the dependency in the headings plugin, import the lib-nekohtml plugin instead. This is done in the build file of the headings plugin.


lewismc (approved):
+1

PR #930, opened 2026-06-16
NUTCH-3188 CI builds startup failure because of disallowed workflow actions

sebastian-nagel:
Update dorny/paths-filter following [ASF approved GH workflow action patterns](https://github.com/apache/infrastructure-actions/blob/main/approved_patterns.yml).


lewismc (approved):
+1 thanks for catching this. It happens every now and then. It's a moving target.

PR #940, opened 2026-07-07
NUTCH-3186 Split SonarQube workflow in unprivileged build and privileged analysis part

sebastian-nagel:
- Cache / upload in master / PR ci build:
  - build artifacts (class files and dependency jars)
  - coverage reports
- Do not run builds or tests in "sonarcloud" workflow
- Download cached artifacts and call Sonarcloud scan
- Do not clean before running integration tests (protocol and indexer) to avoid that test results are removed


lewismc:
Nice @sebastian-nagel

PR #943, opened 2026-07-22
NUTCH-3191 Fix SonarCloud PR analysis: missing JaCoCo XML, test binaries path, and source encoding warnings

lewismc:
PR for [NUTCH-3191](https://issues.apache.org/jira/browse/NUTCH-3191), context in that ticket.

### DIFF
--- a/.github/workflows/junit-report.yml
+++ b/.github/workflows/junit-report.yml
@@ -55,13 +55,15 @@ jobs:
             echo "::error::Download path $root is missing."
             exit 1
           fi
-          files=("$root"/build/**/TEST-*.xml)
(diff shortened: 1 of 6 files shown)


lewismc:
Quick heads up, I created a new SonarCloud Quality Gate profile for us named "Nutch Way" which implements all of Sonar's best practices with the exception that coverage on new code is reduced from 80% to 0%. Reason being, imho 80% was simply too high. We never had project policy so 80% was superfluous anyway.

sebastian-nagel (approved):
Thanks, @lewismc! Looks good.

PR #944, opened 2026-07-24
NUTCH-3192 Address Yetus detsecrets/blanks and SonarCloud security findings in config templates and DummySSL

lewismc:
Fix for [NUTCH-3192](https://issues.apache.org/jira/browse/NUTCH-3192)
- Replace shipped RabbitMQ URI defaults with `amqp://localhost:5672/` (index-writers template, nutch-default `rabbitmq.publisher.server.uri`, indexer-rabbit README); document that a typical local broker uses the well-known guest/guest account and that operators must supply `username:password@` when required
- Change DummySSL SSLContext protocol from "`SSL`" to "`TLS`" (keep intentional trust-all behavior for protocol-httpclient). Do we need to rename the Method and Class?
- Strip trailing whitespace in `httpclient-auth.xml.template` and replace example credentials with `YOUR_USERNAME` / `YOUR_PASSWORD` placeholders
- Mark S6736 as false positive in SonarCloud (masking regex in describe())

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -5,6 +5,7 @@ MD003: false
 MD012: false
 MD013:
   line_length: 200
+  tables: false
 MD022: false
 MD025: false
 MD033: false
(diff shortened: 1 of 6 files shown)


sebastian-nagel (approved):
Thanks, @lewismc. Lgtm!

PR #947, opened 2026-08-05
NUTCH-3196 Yetus: stop failing the tree on known false positives and non-source noise

lewismc:
PR for [NUTCH-3196](https://issues.apache.org/jira/browse/NUTCH-3196)


sebastian-nagel (approved):
Thanks, @lewismc! Looks good.

Very nice to reduce the noise of false positives. And great to see a few more typos eliminated.

PR #949, opened 2026-08-05
NUTCH-3201 Fetcher launches only a single reduce task, ignoring the configuration

sebastian-nagel:
Remove the instruction which configures the single reduce task.


lewismc:
Thanks @sebastian-nagel damn I remember adding this to profile locally and neglected to clean it up. Fantastic catch. This is just a bad bug. Using something like [bats](https://github.com/bats-core/bats-core) to unit test `nutch` and/or `crawl` script wouldn't have caught it either. I kinda wish MRUnit wasn't sent to the attic as it could have been used to test such cases. Anyway +1 from me.

lewismc (approved):
+1

sebastian-nagel:
Thanks, Lewis. Indeed, MRUnit was useful.

PR #954, opened 2026-08-15
NUTCH-3206 XML parsers should not be vulnerable to XXE attacks

lewismc:
PR for [NUTCH-3206](https://issues.apache.org/jira/browse/NUTCH-3206) which is flagged by https://github.com/apache/nutch/security/code-scanning/24
This is a really straightforward PR and should likely have been addressed a while back when a similar concern was reported and addressed.


sebastian-nagel (approved):
Looks good to me. Thanks, @lewismc!

PR #955, opened 2026-08-16
NUTCH-3207 Unify indentation in nutch-default.xml

sebastian-nagel:
A white-space only change (except a single added comma).

Unifies the indentation of every property to 2 spaces. For example:
```xml
<property>
  <name>http.timeout</name>
  <value>10000</value>
  <description>
    The default network timeout, in milliseconds.
  </description>
</property>
```

The description is always kept in a separate block of lines between the opening `<description>` and the closing `</description>`.

### DIFF
--- a/conf/nutch-default.xml
+++ b/conf/nutch-default.xml
@@ -27,20 +27,24 @@
 <property>
   <name>file.content.limit</name>
   <value>1048576</value>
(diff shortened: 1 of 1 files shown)


lewismc (approved):
+1

PR #958, opened 2026-08-16
NUTCH-3197 Yetus: fix real precommit defects (shell, Docker, shelldocs)

lewismc:
PR for [NUTCH-3197](https://issues.apache.org/jira/browse/NUTCH-3197)

Follow-up to [NUTCH-3196](https://github.com/apache/nutch/pull/947) (Child A: Yetus false-positive baselines). This is **Child B** of the Yetus master precommit initiative: fix real defects the gate correctly reported, without taking on historical style debt (Child C) or shellcheck warnings/notes (Child D).

**Fork:** `lewismc/nutch` → `apache/nutch` (replaces incorrectly opened #956 from upstream branch).

## Summary

- Fix 11 shellcheck **errors** in `src/bin/crawl` and `src/bin/nutch` (SC2071, SC2144, SC2242, SC2145) with behavior-preserving edits
- Fix SC2006 on cygpath line introduced by the SC2144 refactor (`$(...)` instead of backticks)
- Add `@audience` / `@stability` shelldocs annotations to five helper functions in `src/bin/crawl`
- Restructure `docker/Dockerfile` for hadolint (combined RUNs, quoting); waive DL3018 via `.hadolint.yaml`
- Exclude `src/plugin/parse-js/sample/` from jshint via `.yetus/excludes.txt` (test fixture, not production JS)

Product **codespell** typos are in a linked follow-up PR from the same fork.

## Out of scope (sibling issues)

- Historical blanks/tabs, `@author`, markdownlint → Child C (NUTCH-3198)
- Shellcheck warnings/notes in bin scripts → Child D (NUTCH-3199)
- Product codespell typos → linked PR from `NUTCH-3197-codespell`

## Test plan

- [x] `ant clean runtime test`
- [x] `shellcheck -S error src/bin/crawl src/bin/nutch` — no errors
- [x] Yetus CI clears shellcheck, shelldocs, hadolint, jshint on changed lines

### DIFF
--- a/src/bin/nutch
+++ b/src/bin/nutch
@@ -147,17 +147,18 @@ fi
 
 local=true
 
-# NUTCH_JOB 
-if [ -f "${NUTCH_HOME}"/*nutch*.job ]; then
-  local=false
-  for f in "$NUTCH_HOME"/*nutch*.job; do
+# NUTCH_JOB
+for f in "${NUTCH_HOME}"/*nutch*.job; do
+  if [ -f "$f" ]; then
+    local=false
     NUTCH_JOB="$f"
-  done
-  # cygwin path translation
-  if $cygwin; then
-	NUTCH_JOB="`cygpath -p -w "$NUTCH_JOB"`"
+    # cygwin path translation
+    if $cygwin; then
+      NUTCH_JOB="`cygpath -p -w "$NUTCH_JOB"`"
+    fi
+    break
   fi
-fi
+done
 
 JAVA="$JAVA_HOME/bin/java"
 JAVA_HEAP_MAX=-Xmx4096m
@@ -238,7 +239,7 @@ fi
 # figure out which class to run
 if [ "$COMMAND" = "crawl" ] ; then
   echo "Command $COMMAND is deprecated, please use bin/crawl instead"
-  exit -1
+  exit 1
 elif [ "$COMMAND" = "inject" ] ; then
   CLASS=org.apache.nutch.crawl.Injector
 elif [ "$COMMAND" = "generate" ] ; then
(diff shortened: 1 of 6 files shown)


src/bin/nutch
```diff
@@ -147,17 +147,18 @@ fi
 
 local=true
 
-# NUTCH_JOB 
-if [ -f "${NUTCH_HOME}"/*nutch*.job ]; then
-  local=false
-  for f in "$NUTCH_HOME"/*nutch*.job; do
+# NUTCH_JOB
+for f in "${NUTCH_HOME}"/*nutch*.job; do
```
sebastian-nagel, line 151: If there are two or more job files in NUTCH_HOME, that's already an issue very likely. But ok for this fix. If we are lucky the last job file is the right one.

src/bin/nutch
```diff
@@ -336,7 +337,7 @@ else
  # check that hadoop can be found on the path
  if [ $(which hadoop | wc -l ) -eq 0 ]; then
     echo "Can't find Hadoop executable. Add HADOOP_COMMON_HOME/bin to the path or run in local mode."
-    exit -1;
```
sebastian-nagel, line 339: This is somewhat sensitive because the script bin/crawl expects that `bin/nutch generate ...` exits with code 1 when no segment was created. All errors should ideally use other error codes.

Of course, this is not 100% working, because the Generator uses Hadoop's ToolRunner which maps exceptions (don't whether all) to 1.

sebastian-nagel: Maybe we should change the behavior of Generator and exit with a more unique code, in case no segment was created (nothing to fetch). But would be a breaking change.

sebastian-nagel (approved):
Thanks, @lewismc. Looks good.

The exit values may need some extra review / thinking.

src/bin/nutch
lewismc: I agree. More than one *nutch*.job already means a messy NUTCH_HOME. This change only makes the glob shellcheck-safe and takes the first match. Happy to fail closed on multiples in a follow-up if you want that.

src/bin/nutch
lewismc: Yes this was some education for me. `1` would mean “no segment” for bin/crawl + Generator... which is misleading/incorrect.
A missing Hadoop is now exit `127` (POSIX utility-not-found) in `bin/nutch` and `bin/crawl`, so that path is no longer confused with an empty generate. `ToolRunner` still mapping some exceptions to `1` is unchanged.

lewismc:
Hi @sebastian-nagel thanks for the review. In summary, this PR now keeps generate’s `1`. Env/Hadoop-not-found is `127`.

A dedicated empty-generate code can be its own JIRA if we want to break that contract... I didn't bother here.

I'll leave this open for a while longer, if you want to revisit. Thanks.

sebastian-nagel:
Thanks, @lewismc! Please merge!

PR #962, opened 2026-08-31
NUTCH-3208 GitHub Actions JUnit and Yetus workflows do not reliably post PR comments

lewismc:
(junit): post a sticky totals comment when parent PR CI succeeds or fails and XML is present. Stay silent if tests were path-filtered.

PR fixes part 1 of [NUTCH-3208](https://issues.apache.org/jira/browse/NUTCH-3208)

Unfortunately, the GitHub workflow result cannot be properly validated unless it is merged into `master` and then the workflow is run on a new incoming PR...which is what I did in my local setup.

### DIFF
--- a/.github/workflows/junit-report.yml
+++ b/.github/workflows/junit-report.yml
@@ -31,7 +31,10 @@ permissions:
   pull-requests: write
 jobs:
   checks:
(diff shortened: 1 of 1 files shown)


sebastian-nagel (approved):
Thanks, @lewismc.

PR #967, opened 2026-09-19
NUTCH-3130 Address deprecated API usage across Nutch codebase and build

lewismc:
Updated PR for [NUTCH-3130](https://issues.apache.org/jira/browse/NUTCH-3130) which replaces stale https://github.com/apache/nutch/pull/869

#869 had drifted: it was branched before master moved to Java 17 only and before the REST service was removed, so a large part of its diff was either already applied or no longer applied cleanly. Rather than rebase a stale branch with a long merge history, this is a fresh branch against current master containing only the work that still matters.

The scope is deliberately narrower than #869. Two changes there were never settled in review and are left out: removing `finalize()` from `Plugin`, `PluginRepository` and `Ftp` (neither @sebastian-nagel nor I could show that plugin and FTP shutdown still happen without it), and swapping `SpellCheckedMetadata` for `CaseInsensitiveMetadata` in the protocol plugins, which changes HTTP header handling and belongs with NUTCH-3002. Both should be follow-up issues.

What remains is the mechanical work plus the CI gate. Deprecated API calls are replaced: commons-cli `OptionBuilder`/`GnuParser` become `Option.builder` and `DefaultParser`, `GeoIPDocumentCreator` uses the geoip2 5.x accessors, and `SpellCheckedMetadata` uses commons-text for Levenshtein distance with the same threshold and the same behaviour. Any deprecated Nutch API that is unused or long past it's removal date is deleted: `IndexWriter.open(Configuration, String)` (deprecated since 1.15) and its plugin stubs, the old `net.protocols.ProtocolException`, and two unused overloads in `Generator` and `CommonCrawlFormatFactory`. Deprecated API's we are keeping now says in Javadoc what to use instead.

The "Check for deprecation warnings" step in the GitHub Actions workflow from #869 is kept, with two fixes: it skips cleanly when no test compile ran instead of reporting a green pass, and the allowlist is rebuilt from what actually still warns on master rather than being carried over verbatim.

Removing `IndexWriter.open(Configuration, String)` is a breaking change for out-of-tree index writer plugins, which must implement only `open(IndexWriterParams)`. Everything else is reversible. We can call this out in the 1.24 release notes.


lewismc:
Follow-up after first CI runs...

My commons-cli rewrite copied `OptionBuilder` method names onto Option.Builder. `hasOptionalArg()` and `hasArgs(2)` do not exist on the builder API; NodeDumper and NodeReader now use `optionalArg(true)` and `numberOfArgs(2)`.

Ivy was resolving a newer commons-cli than Hadoop 3.5.0 ships (1.9.0), so javac warned that `HelpFormatter` and `Option.Builder.build()` were deprecated (1.10+ APIs). Switching to those APIs would break on a Hadoop 3.5 classpath. Core `ivy/ivy.xml` now pins `commons-cli:commons-cli:1.9.0` with `force="true"`. Those warnings are gone.

`FetchItemQueues` uses Guava `expireAfterWrite(Duration)` instead of the deprecated (long, TimeUnit) overload.

The deprecation allowlist is now only `Plugin`, `PluginRepository`, `Ftp` (finalize(), still deferred) and the two Hadoop `JobContext` test stubs. The gate runs on JDK 17 only; JDK 21 javac emits java.net.URL constructor warnings that are out of scope here.

I updatred protocol-ftp Client uses Commons Net `FTPCmd` / `sendCommand(FTPCmd, String)` instead of deprecated `FTPCommand` ints.  The`plugin.xml` lists `commons-net-3.9.0.jar` to match that plugin’s Ivy revision.

lewismc:
I'll address the SonarCloud issues in a follow-on PR. They are to do with the docker image analysis.

src/java/org/apache/nutch/plugin/PluginRepository.java
```diff
@@ -98,7 +98,7 @@ public PluginRepository(Configuration conf) throws RuntimeException {
     try {
       installExtensions(this.fRegisteredPlugins);
     } catch (PluginRuntimeException e) {
-      LOG.error("Could not install extensions.", e.toString());
+      LOG.error("Could not install extensions: {}", e.toString());
```
sebastian-nagel, line 101: +1

sebastian-nagel (approved):
Thanks, @lewismc. Looks good!

